PR #1, opened 2022-02-25
2685 add basic styling

adamjarling:
This PR does a few things including:
- Adds some basic styling to the pages
- Supports grabbing data from Collection List, to an individual Collection (displays Collection Works), and then finally Work data.  Everything is pulled from our Elasticsearch Staging index.
- All data and routes are created at build time.  Currently pre-builds 10,000 Work routes and pages.  We can play with this to ramp up the number and see how long builds take.

Next Steps:
- Use Next's `/pages/api/` setup to start authoring a more client-friendly API, delivering minimal data DC (or any client) may need.

@mathewjordan I'd like to keep all these commits and not squash, in case we want to trace back and reference code as there are some different approaches which might be useful.


pages/collection/list.tsx
```diff
@@ -18,21 +20,25 @@ interface CollectionListProps {
   collections: Array<Collection>;
 }
 
-const CollectionList: React.FC<CollectionListProps> = ({ collections }) => {
+const CollectionList: NextPage<CollectionListProps> = ({ collections }) => {
```
mathewjordan, line 23: Love this next types, I'm going to read up on all of them that might be available to us.

mathewjordan (approved):
I think this is great. Thanks for bringing over my stitches stuff.

PR #2, opened 2022-03-10
Create baseline Facet component with types

mathewjordan:
# What does this do?

This PR just organizes things a bit more and breaks some our logic into a Facet component. I think it might be best to complete quite a few micro PRs rather than trying to merge everything after all of our individual exploration. 

# How do I review this?

Things should functionally with the same as they did in [2710-api-connect-testing](https://github.com/nulib/dc-nextjs/tree/2710-api-connect-testing). Running `npm run dev` should spins things up and checking any facet options should `console.log()` each individual event.

### DIFF
--- a/components/Facet/Facet.tsx
+++ b/components/Facet/Facet.tsx
@@ -7,13 +7,10 @@ interface FacetBuckets {
 export interface FacetProps {
   label: string;
   buckets: FacetBuckets[];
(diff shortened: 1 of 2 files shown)


adamjarling (approved):
Looking good

kdid (approved):

PR #5, opened 2022-03-11
Add search constants to build aggs.

mathewjordan:
This just adds some data for facets that will help us programmatically build aggs on the elastic search requests. I don't have my heart set on anything here other than that we should somehow dynamically generate this in some way.

From this you should now see all the Facets populating on the search page.


adamjarling (approved):
Looks good.  Makes sense.

PR #6, opened 2022-03-11
Quick cleanup of results bug, and small refactor to search component …

adamjarling:
This cleans up a logic a little in the Search Page component, and fixes a bug in re-creating the ES Query from the template helper.  Previously it was holding onto previously selected facet values.


mathewjordan (approved):
Think you'll need to update the size in the new aggs.ts but otherwise looks good.

PR #13, opened 2022-03-28
Initial progress toward facet filters

kdid:
Gets us started towards facet filters 
 - Adds the filter query
 - Shows filter results above regular facet results
 - Excludes any selected values from the filter results
 - (* Other than this^ doesn't handle much of the interaction with the facet or rest of the search.)

Screen Shot 2022-03-28 at 8 48 37 AM

### DIFF
--- a/components/FacetFilter/FacetFilter.tsx
+++ b/components/FacetFilter/FacetFilter.tsx
@@ -1,0 +1,20 @@
+export interface FacetFilterProps {
+  facetLabel: string;
+  handleFacetFilterChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+const FacetFilter: React.FC<FacetFilterProps> = ({
+  facetLabel,
+  handleFacetFilterChange,
+}) => {
+  return (
+    <input
+      type="text"
+      id={`${facetLabel}-filter`}
+      name={facetLabel}
+      onChange={handleFacetFilterChange}
+    />
+  );
+};
+
+export default FacetFilter;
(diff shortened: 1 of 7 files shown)


components/FacetFilter/FacetFilter.tsx
```diff
@@ -0,0 +1,20 @@
+export interface FacetFilterProps {
```
adamjarling, line 1: If we're not using `FacetFilterProps` in any other components, could lose the `export` here?  Typically `interface`s which have "Props" in the name are generally inferred to be local to the component (at least the way we've been doing things).

lib/queries/facet-filter.ts
```diff
@@ -0,0 +1,132 @@
+import { UserFacets } from "types";
+import { FacetsInstance, ALL_FACETS } from "lib/facets";
+
+const escapeRegExp = (string: string) => {
+  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+};
+
+const facetFilter = (facetLabel: string, facets: Array<FacetsInstance>) => {
+  return facets.find(({ id }) => id === facetLabel)?.field;
```
adamjarling, line 9: Nice one-liner!

pages/search.tsx
```diff
@@ -65,9 +68,42 @@ const SearchPage: NextPage = () => {
...
+        ),
+      })
+        .then((response) => {
+          return response.json();
+        })
+        .then((facetFilterData) => {
+          const newObj: FilteredFacets = { ...facetFilterResults };
+          newObj[name] = facetFilterData.aggregations.facetFilter.buckets;
+          setFacetFilterResults(newObj);
+          return;
```
adamjarling, line 102: Probably don't need this `return`, as another function doesn't seem to be waiting for a resolved Promise?

adamjarling (approved):
This looks great for testing out how this would work.  We can clean it up and refine it as we dive back in.

PR #15, opened 2022-04-07
2853 set up linting testing

adamjarling:
This PR wires up ESLint and Jest.  For ESLint, I'd like to be pretty opinionated about Code Quality in this project to run a tight ship.  You'll notice I added in the ESLint config file stuff like alphanumeric order checking for `@imports` and `Object key` organization.  They're all throwing warnings now to keep the build moving, but as we work through stuff I'd like to fix up what we can when in each file.

There's also a pattern in place for relative imports that works with both our app and testing.  The convention is:

```jsx
import Figure from "@/components/Figure/Figure";
import { someRootFile} from "@/probably-stitches-here";
...
```

Take a look at our `tsconfig.json` and `jest.config.js` for which aliases are being used, or to add new ones.

### To Test
- Pull down the branch
- `npm i`
- `npm lint`
- `npm test`

To see how it works and looks.

### DIFF
--- a/components/Figure/Figure.styled.ts
+++ b/components/Figure/Figure.styled.ts
@@ -1,4 +1,4 @@
-import { styled } from "stitches.config";
+import { styled } from "@/stitches.config";
 
 const FigureStyled = styled("figure", {
   display: "flex",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,26 @@
 {
-  "extends": ["next/core-web-vitals", "prettier"]
+  "parser": "@typescript-eslint/parser",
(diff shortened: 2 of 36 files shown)


components/Figure/Figure.styled.ts
```diff
@@ -1,4 +1,4 @@
-import { styled } from "stitches.config";
+import { styled } from "@/stitches.config";
```
mathewjordan, line 1: Love this relative import.

mathewjordan (approved):
Good to go. 🚀 Makes sense to get this covered before we really start moving heavy on this.

PR #17, opened 2022-04-12
Address typescript error in facet filter query

kdid:
Build error

```
[WARNING]: ./lib/queries/facet-filter.ts:18:38
                                    Type error: Argument of type 'string' is not assignable to parameter of type 'FacetsInstance[]'.
```


mathewjordan (approved):
Makes sense to me.

PR #19, opened 2022-04-14
Wireup search results using real OpenSearch results data

mathewjordan:
- Blank slate `/search`.
- Mock out Heading component.
- Update type defs for heading elements.
- Wireup Grid to results.
- Load in real results for each GridItem.
- Tidy up search methods.
- Clean house a bit more.
- Remove unused type defs.
- Remove Sticky.
- Wireup oldstyle get.
- Get search results firing a single time.
- Make queries cleaner.
- Refine type defs on search component.
- Populate search bar on router update.
- Carry searchValue when updating router.
- Finish up basic search tie ins.
- Remove useState in test.

@kdid This PR basically trims down the search page, removing facets temporarily, and integrates in the Search component as the header. To test, I setup some basic next routing that carries the query string to a URL pattern of /search?q={string}` with the `q` as a GET.  When @adamjarling is back, we'll want to dive into how we wireup storing that search param globally. 

I also have tests working locally now, so I'd like to see if this works in amplify as well.

To test:

- `npm run dev`
- browse to http://localhost:3000/
- input any string into the search bar, ex: `dog`, `joan`
- actual results  should populate in the Grid (I've disabled thumbnails)
- hit the (X) clear button in search bar, this should reset the query to default

## Notes
 - Things are expected to be unstyled right now so don't give that too much consideration.
 - All results up the limit should populate right now, paging/continuous scroll will handle breaking that out later.
 - This is still conforming to the elastic search response shape right now, we'll just need to update that at some point.


kdid:
@mathewjordan - I think this is going in a good direction by simplifying/starting fresh. The code looks good (to me). I'm a bit confused by what's happening with the firing of the search request to the API. It seems like if I type something like "dog" in the search bar, the search is executing three times - two times w/out the query string and one time with. Is there a reason it's doing this?

mathewjordan:
> @mathewjordan - I think this is going in a good direction by simplifying/starting fresh. The code looks good (to me). I'm a bit confused by what's happening with the firing of the search request to the API. It seems like if I type something like "dog" in the search bar and hit enter, the search is executing three times - two times w/out the query string and one time with. Is there a reason it's doing this?

Thanks @kdid. I took a deeper look a this and now have whittled down just two requests on initial `/search` page load. Subsequent queries only have one pair of a request/response. All signs point towards this being related to my use of `useRouter()` in Next to get the query var. This is causing the `q` to shift in value. I think we will want to dig into this deeper as a micro issue as I imagine we will want to lean on Next for this type of behavior and not use an older school approach to get this value which might be an anti pattern in Next.

https://nextjs.org/docs/advanced-features/automatic-static-optimization#how-it-works

> During prerendering, the router's query object will be empty since we do not have query information to provide during this phase. After hydration, Next.js will trigger an update to your application to provide the route parameters in the query object.

kdid (approved):

mathewjordan:
Thanks @kdid.

I created https://github.com/nulib/repodev_planning_and_docs/issues/2866

PR #20, opened 2022-04-15
Refresh Header, minimally Northwesternize some stuff.

mathewjordan:
image

## What does this do?

This PR further cleans up the styling of the Header and gets us moving in the right direction as far as brand goes. I've hardcoded in a limit of `20` to test look and feel with actual thumbnails in responses.

- Minimally Northwesternize, applying color
- Style container `div` elements to extend to Northwesternize size widths (grid extends to the Wide width).
- Updates the search bar, and adds the **Browse Collections** menu item

## Review

Please visit http://localhost:3000 after running `npm run dev`. Input a search for anything if you'd like to test how his looks with actual results.


kdid (approved):
Looks good

PR #24, opened 2022-04-27
Initiate Facet component.

mathewjordan:
- Prep facet filter component layout.
- More prep work to pass data to facets.
- Add tests and some minimal styling for facets.
- Minimal styling.
- Breakout Options component.
- Update tests, cleanup component types.
- 
@adamjarling This begins to breakout the Facet component tray that will render on the search page. The logic works like this:

1. `<Facets/>` intakes aggregations from the API data response on the search page
3. Aggregations are sent down to `<Filter/>` component containing a body that will render out the modal AND breaks off the worktype aggregation and renders out an `<InlineFacet/>` justified right. The Inline facet is just a placeholder for now but should behave like other facets like components.
4. All Facets mapped to the modal render as a `<MultiFacet/>` which props in the shape of an **ApiResponseAggregation**
5. The `<MultiFacet/>` maps the buckets contained within it as `<Options/>` that will be rendered as list item checkboxes

Note: None of the work here interacts with our react context. That will be undertaken in a future PR.


components/Facets/MultiFacet.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { render, screen, within } from "@testing-library/react";
+import MultiFacet from "./MultiFacet";
+import React from "react";
+import { mockAggregation } from "@/mocks/aggregation";
+
+describe("MultiFacet UI component", () => {
+  it("Renders a multi-select facet component.", async () => {
+    render(<MultiFacet {...mockAggregation} />);
```
adamjarling, line 8: One trick for this `render` repeated, is to put in in a 
```
beforeEach(() => {
  render(<MultiFacet {...mockAggregation} />);
}) 
```
which lives directly under the `describe` line and before the first `it`.  Then each spec should have access to it.

If you're changing up the props passed in for each spec, this isn't always the best solution, but saves a few lines of code.

components/Facets/Option.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const mockOption: FacetOption = {
+  bucket: mockAggregation.buckets[0],
+  facet: mockAggregation.id,
+  index: 0,
+  type: "checkbox",
+};
+
+describe("Facet Option UI component", () => {
+  it("Renders a facet option as a checkbox.", async () => {
```
adamjarling, line 15: And here I think you could lose the `async` and in the spec instead of using `getByRole`, use `findByRole` which an asynchronous selector.   Check out the testing-library docs for more info, but it's pretty handy.

adamjarling (approved):
Lot of work here and looks great.  A few suggestions on tests (thanks for adding 'em), and looks like a conflict to resolve after my latest merge.  We're rocking!

mathewjordan:
Thanks @adamjarling, I'll tidy those tests up, resolve conflicts and push through.

components/Facets/MultiFacet.test.tsx
mathewjordan: Looks like our linting does not like this approach. 
image

PR #25, opened 2022-04-28
Add test-utils helper to fix failing Jest tests

adamjarling:
Add test-utils helper to optionally wrap components with our SearchProvider.

See any of the tests for an implementation.  If we want to provide a unique, custom `context` to any of our tests, we can manually import the `<SearchProvider />` component into our test, and feed it an initial state(ie. `<SearchProvider initialState={ { q: "foo", userFacets: { ... } } />`)

### DIFF
--- a/test-utils.tsx
+++ b/test-utils.tsx
@@ -1,0 +1,15 @@
+import React, { FC, ReactElement } from "react";
+import { RenderOptions, render, screen } from "@testing-library/react";
+import { SearchProvider } from "@/context/search-context";
+
+const AllTheProviders: FC = ({ children }) => {
+  return <SearchProvider>{children}</SearchProvider>;
+};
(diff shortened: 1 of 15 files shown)


test-utils.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return <SearchProvider>{children}</SearchProvider>;
+};
+
+const customRender = (
+  ui: ReactElement,
+  options?: Omit<RenderOptions, "wrapper">
+) => render(ui, { wrapper: AllTheProviders, ...options });
+
+export * from "@testing-library/react";
+export { customRender as render, screen };
```
mathewjordan, line 15: This will be immensely helpful! 👍

mathewjordan (approved):

PR #26, opened 2022-05-05
Cleanup of Facet filtering code

adamjarling:
This is an attempt to clean up the code, without touching too much of the Facet code you're working on.  

After this gets merged, we should get rid of any code we're not using, including `search-old.tsx`.


mathewjordan (approved):
Thanks, looks good. We'll see if it conflicts with my stuff in a bit.

PR #30, opened 2022-05-13
Wire up initial data flow in the Filter modal component

adamjarling:
This PR wires up some initial data flow in the Filter component.


components/Facets/GroupList.tsx
```diff
@@ -0,0 +1,50 @@
+import * as Accordion from "@radix-ui/react-accordion";
```
mathewjordan, line 1: At some point, we'll want to clean these Radix UI things up.

mathewjordan (approved):
:shipit:

PR #31, opened 2022-05-16
Fix facet component filter display.

mathewjordan:
Fixes small bug that causes the Facets component (and subcomponents) to not render.


mathewjordan:
@adamjarling This is the fix I put in place to help with demoing last Friday. We may want to piece aggs back in once we handle https://github.com/nulib/repodev_planning_and_docs/issues/2819

PR #32, opened 2022-05-16
Wireup work type facet filter.

mathewjordan:
image

### What does this do?
This adds back a Work Type component that is always exposed on the search page.  Because the options **All, Audio, Image, and Work** should always be visible, the facet options are mapped from a const in the `facets-model.ts` file. 

### What does this not include?
- `doc_count` on number of items for each option
- a `disabled` attribute if option `doc_count` is `0`


adamjarling (approved):
Yea, think this whole thing makes sense and some nice, slick styling to boot.  Good work.

PR #33, opened 2022-05-17
Add support for Facet agg values filtering

adamjarling:
Pretty rough around the edges, but curious on your thoughts on the general approach.


components/Facets/FacetWrapper.tsx
```diff
@@ -39,18 +41,33 @@ const FacetWrapper: React.FC<FacetWrapperProps> = ({ facet }) => {
...
-
-  return <>{fileredAggregation}</>;
+    .map((aggregation) => {
+      const userBuckets = userFacetsAggregation
+        ? userFacetsAggregation.buckets
+        : [];
+      const filteredAggBuckets = aggregation.buckets.filter((ab) => {
+        return !userBuckets.find((ub) => ub.key === ab.key);
+      });
+      const buckets = [...userBuckets, ...filteredAggBuckets];
```
mathewjordan, line 57: Maybe we could also push this out to a hook/lib service to keep things tidy? Also could be a separate optimization issue.

hooks/useFetch.ts
```diff
@@ -0,0 +1,53 @@
...
+  error: ApiError;
+  loading: boolean;
+};
+
+type HookArguments = {
+  body?: ApiSearchRequest;
+  url: string;
+};
+
+const useFetch = (obj: HookArguments): Response => {
```
mathewjordan, line 19: I like the idea of keeping our api fetch hook pretty generic and ready go.

lib/queries/facet-filter.ts
```diff
@@ -51,13 +56,13 @@ export const buildFacetFilterQuery = (
...
 
-const baseQuery = (facetFilterKey: FacetFilterKey, term: string) => {
+const baseQuery = (facetFilterKey: FacetFilterKey, filterTerm: string) => {
   return {
     aggs: {
       facetFilter: {
         terms: {
           field: `${facetFilterKey}`,
-          include: `.*${term}.*`,
+          include: `.*${filterTerm}.*`,
```
mathewjordan, line 65: We may want to bring this regex to a broader discussion with the team at some point. How specific should our include be? Looks good now.

mathewjordan (approved):
✨ I say, let's get this through and optimize things later. Overall, looks great. I wonder if we should next have a spring cleaning issue to throw away our old junk?

PR #34, opened 2022-05-20
Add active user facets tray to Search and Modal screens.

mathewjordan:
image

## What does this do?

This little gem brings in the UserFacets component, which in turn renders our the `userFacets` and `userFacetsUnsubmitted` from the respective search and filter contexts. Both the default screen for search and the modal are handled from the same component, with some conditional rendering occurring for the instances.

- If `userFacets` or `userFacetsUnsubmitted` are empty objects, the component will render an empty React fragment.
- For the `search` screen, an indicator will render next to a dropdown, informing the user to the count on the number of facet options they have selected.
- Toggling the dropdown on the `search` screen will render column of values that can be clicked.
- Clicking the Filter button directly will still activate the modal. 
- Option values in the `modal` screen will wrap in a div rendered above the facet options. Long term, we may want to clean this up.

## How should we review?
For the most part, reviewing this is straightforward. Some care should probably be taken to review how I handled the using the UserFacets component with multiple contexts. I could image there will be some conflicts and cleanup to move some helper utils out of my components after we have things combined without other concurrent work.

## Notes
- Secondary colors (NON-PURPLE!) have been added to our stitches config.
- I also took advantage of time to reorganize our SVG and Container components into a `Shared` components directory.
- At some point, we may want to EXCLUDE work type from being considered an active facet. I think that might be a larger group decision though.


components/Facets/UserFacets/UserFacets.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { render, screen } from "@testing-library/react";
+import { FilterProvider } from "@/context/filter-context";
+import React from "react";
+import { SearchProvider } from "@/context/search-context";
+import UserFacets from "./UserFacets";
+
+const searchStateDefault = {
+  aggregations: [],
+  q: "",
+  userFacets: { genre: ["Foo"] },
```
adamjarling, line 11: Should this `userFacets` value be `{}`?

components/Facets/UserFacets/UserFacets.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <UserFacets screen="modal" />
+        </FilterProvider>
+      </SearchProvider>
+    );
+    const userFacets = screen.getByTestId(`facet-user-component`);
+    expect(userFacets).toBeInTheDocument();
+    const values = screen.getAllByTestId(`facet-user-value-component`);
+    expect(values.length).toBe(3);
+  });
+});
```
adamjarling, line 71: Tests look great.  One idea to test this functionality could be asking not only does the component render to the DOM, but say for screen, does it render `within` some search page wrapping element (as opposed to rendering inside the Filter markup)?   These tests assume it could render anywhere in the DOM, no?

components/Facets/UserFacets/UserFacets.tsx
```diff
@@ -0,0 +1,128 @@
...
+      );
+      break;
+  }
+
+  useEffect(() => {
+    if (!facets) return;
+
+    const current: CurrentFacet[] = [];
+    for (const [facetId, facetKeys] of Object.entries(facets)) {
+      const facet = ALL_FACETS.facets.find(({ id }) => id === facetId);
```
adamjarling, line 52: Maybe we could use or create one of those nifty little helper functions in `lib/utils/facets-helpers`?

adamjarling (commented):
This is super slick.  A few minor thoughts, but overall good to go.

components/Facets/UserFacets/UserFacets.test.tsx
mathewjordan: @adamjarling This makes sense to me, however wouldn't rendering `within` require me to also mock the wrapping element or page? Really the component is fairly dumb and only renders/returns its layout because of the `screen` prop.

adamjarling (approved):

PR #35, opened 2022-05-20
Handle opening closing of Filter dialog and state management

adamjarling:
## What does this PR do?
This code will handle passing data state between opening and closing the Filter dialog.  From the parent Search Context, down to the Filter Context, and vice versa.  

When the Filter dialog is opened, the current state of User Facets is passed into the Filter dialog.  If the user had previously interacted with the Filter, their last interacted with Facet Group will open by default, and all current user selected facets will be represented in the UI.

When the Filter dialog closes, we pass the state of all selected user facets back up to the Search page.  We then clear out selected user facets from the Filter Context, since the user might manipulate their selected facets while searching.

### DIFF
--- a/components/Facets/Filter.tsx
+++ b/components/Facets/Filter.tsx
@@ -6,13 +6,22 @@ import {
   FilterOverlay,
   FilterTrigger,
 } from "./Filter.styled";
+import { FilterProvider, useFilterState } from "@/context/filter-context";
 import FacetsGroupList from "./GroupList";
-import { FilterProvider } from "@/context/filter-context";
+import FacetsSubmit from "@/components/Facets/Submit";
 import Preview from "./Preview";
+import React from "react";
+import { useSearchState } from "@/context/search-context";
 
-const FilterModal: React.FC = () => {
+export type SetIsModalOpenType = React.Dispatch<React.SetStateAction<boolean>>;
+
+type FilterModalProps = {
+  setIsModalOpen: SetIsModalOpenType;
+};
+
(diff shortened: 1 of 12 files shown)


lib/utils/facet-helpers.ts
```diff
@@ -0,0 +1,10 @@
+import { ALL_FACETS, FACETS } from "@/lib/constants/facets-model";
+import { FacetsGroup, FacetsInstance } from "@/types/components/facets";
+
+export const getFacetById = (id: string): FacetsInstance | undefined => {
+  return ALL_FACETS.facets.find((facet) => facet.id === id);
+};
```
mathewjordan, line 6: This will be pretty helpful. ✅

components/Facets/Filter.tsx
```diff
@@ -25,25 +34,52 @@ const FilterModal: React.FC = () => {
...
+    /**
+     * Opening the Filter dialog, so transfer the updated user facets
+     * from Search Context back into the Filter Context
+     */
+    if (!isModalOpen) {
+      filterDispatch({
+        type: "updateUserFacets",
+        userFacetsUnsubmitted: searchState.userFacets,
+      });
+    }
```
mathewjordan, line 58: 👍 love the simplicity of the overwriting of this.

mathewjordan (approved):
Looks good, rebase/merge away. :shipit:

components/Facets/Filter.tsx
adamjarling: Cool. An argument still could be made for combining the 2 contexts, but might give us some wiggle room in the future to keep 'em separated.

PR #36, opened 2022-05-20
Update default open state behavior for Filter dialog

adamjarling:
## What does this PR do?
- Updates the default tab/accordion open behavior.  We now save whichever facet the user last selected (not deselected).  And that will open by default when the Filter dialog opens.  Not sure this is the best approach, but we'll have to run it by @davidschober or users.
- Adds more thorough test coverage to `FacetsGroupList` component.


context/filter-context.tsx
```diff
@@ -22,7 +22,7 @@ type FilterProviderProps = {
 };
 
 const defaultState: FilterContextStore = {
-  lastFacetViewed: undefined,
+  recentFacet: undefined,
```
mathewjordan, line 25: I like the rename here.

mathewjordan (approved):
This looks pretty thorough. Now we need to get our other tests up to speed. 🚗 ⚡

PR #38, opened 2022-05-23
Refine filter modal layout and styling.

mathewjordan:
image

## What does this do?
This adds some design structure our modal layout and cleans breaks out the `<Filter/>` component a bit more, by adding a new `<Modal/>` component that renders active content of the Radix Dialog when `open`.

- Some minimal responsiveness of modal layout is accounted for on `sm` screens.
- A cancel button is now visible in the footer
- The **View Results** submit button will display a count total if it exists
- A string will render in the header if a `q` exists on the searchState, if q is `joan baez`, it will say _"Results for **"joan baez"**_
- This also media query breakpoints as `media.ts` in the styles directory.

## What's not included.
Any styling of the GroupList and Preview components


components/Facets/Filter/Filter.styled.ts
```diff
@@ -77,40 +79,153 @@ const FilterOverlay = styled(Dialog.Overlay, {
...
+  position: "absolute",
+  bottom: "0",
+  padding: "1rem",
+  backgroundColor: "$black10",
+  width: "100%",
+  boxShadow: "inset 3px 3px 8px #0001",
+  display: "flex",
+  justifyContent: "flex-end",
+
+  button: {
```
adamjarling, line 119: Looks great.  Curious how we're going to keep all these styles consistent for common elements, for example button here, or are you not worrying about that yet?

components/Facets/Filter/Modal.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      expect(button.getAttribute("aria-label")).toBe("Cancel");
+    });
+  });
+
+  it("Has a visible submit button with a total.", () => {
+    renderHelper();
+    const submit = screen.getByTestId("facets-submit");
+    expect(submit).toHaveTextContent("View Results (20)");
+  });
+});
```
adamjarling, line 46: Nice test structure here

styles/media.ts
```diff
@@ -0,0 +1,12 @@
...
+const media = {
+  xxs: "(max-width: 349px)",
+  xs: "(max-width: 575px)",
+  sm: "(max-width: 767px)",
+  md: "(max-width: 991px)",
+  xl: "(max-width: 1199px)",
+  lg: "(min-width: 1200px)",
+};
+
+export default media;
```
adamjarling, line 12: Sweet.

adamjarling (approved):
Looks good, but see comment about if/should we be thinking about default styles and/or components.  This is on my mind as I'm laying out the Work page.  Should a `h2` be defined somewhere for example, or do we create a `<Headline />` component or something, etc...

components/Facets/Filter/Filter.styled.ts
mathewjordan: We should probably start thinking about that very soon, perhaps make a UI directory of some sort that handles common components and their styling for anything not design system.

PR #39, opened 2022-05-25
Work page placeholder 

adamjarling:
## What does this do?

The Work page placeholder is now in place.  For incremental static regeneration (curious how this works when ends up in AWS), we're grabbing 100 works which will be statically generated at build time.  The other Work pages should dynamically build and be served ok.  We'll test it.

NOTE: there is a test warning that happens because of how we're importing a dynamic component (Clover IIIF, which pulls in OpenSeadragon, which only lives in the browser environment).  Will come back to how to mock either Clover or Openseadragon with a fresh head, or pair with @mathewjordan to sort it out.

### To test/view
Go to the Search page, and click on any work search result image.  

image


components/Work/TopInfo.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const { descriptions, title } = work;
+  return (
+    <TopInfoWrapper>
+      <div data-testid="work-top-info-wrapper">
+        <h1>{title}</h1>
+        {descriptions.length === 0 && <p>No description exists</p>}
+        {descriptions.map((d) => (
+          <p key={d}>{d}</p>
+        ))}
```
mathewjordan, line 24: Title (label), Metadata (metadata), Description (summary)... etc will be Nectar, long-term, yeah? Do we want to bring that in now?

components/Shared/Cart.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render, screen } from "@/test-utils";
```
mathewjordan, line 1: Rename this `Card.test.tsx`?

pages/works/[id].tsx
```diff
@@ -1,14 +1,49 @@
...
+  if (!work) {
+    return null;
+  }
   return (
     <Layout>
-      <Container>Placeholder for single work</Container>
+      <ErrorBoundary FallbackComponent={ErrorFallback}>
+        <WorkViewerWrapper manifestId={work.iiif_manifest} />
+        <Container>
+          <WorkTopInfo work={work} />
```
Comment (mathewjordan) [inline pages/works/[id].tsx:24]: Do we want to hand this `work` or the fetched response of `work.iiif_manifest`?

mathewjordan (changes requested):
Looks like a good start. My main questions are regarding how far we want to take this now. Do we want to do any of the manifest fetching and distilling down to Nectar in this spint? The only requested change is on the type for `Shared/Cart.test.tsx`

components/Shared/Cart.test.tsx
adamjarling: This is now fixed in the latest push

components/Shared/Card.tsx
```diff
@@ -1,27 +1,30 @@
...
-      <CardMetadata>
-        {metadata?.map((meta) => (
-          <span key={meta}>{meta}</span>
-        ))}
-      </CardMetadata>
-      <p>{description}</p>
+      <h3 data-testid="card-title">{title}</h3>
+      <Figure data={data} />
+      <p data-testid="card-description">{description}</p>
     </div>
```
mathewjordan, line 28: Yeah, I like this better. Cards having Figures that is.

components/Figure/Figure.tsx
```diff
@@ -1,21 +1,21 @@
...
 
 const Figure: React.FC<FigureProps> = ({ data }) => {
-  const { title, type, src } = data;
+  const { title, supplementalInfo, src } = data;
   return (
     <FigureStyled>
       <Image src={src} style={{ width: "100%" }} alt={title} />
       <Title>{title}</Title>
-      <Type>{type}</Type>
+      <SupplementalInfo>{supplementalInfo}</SupplementalInfo>
```
mathewjordan, line 18: This maybe should be: 
```jsx
{supplementalInfo && <SupplementalInfo>{supplementalInfo}</SupplementalInfo>}
```

mathewjordan (approved):
I think this is good. Take or leave my suggestion on the `supplementalInfo` in Figure.

components/Work/TopInfo.tsx
adamjarling: Pulling in Nectar in a new issue as chatted about.

pages/works/[id].tsx
Comment (adamjarling) [reply pages/works/[id].tsx:24]: Yea, to be handled in a forthcoming issue, how to grab the IIIF data, where to store it and pass it into any IIIF aware components.

PR #40, opened 2022-05-26
Style modal Group Lists.

mathewjordan:
## What does this do?

This PR styles the Radix UI Accordion and Tab functionality on an open modal following guidance from our wireframes. This styling wraps the Options and GenericFacet components and these are also giving a nice styling treatment.

## What's not included?

This is pretty much just an issue focused on CSS styling. At some point we may also want to consider count on number of options and adding a "Show More" button. We may also want to come back and refine the internal scrolling of the modal body.

image

image

- Initiate group list styling.
- Style Tabs.
- Get facet label by id.
- Style options.
- Adjust doc count test.


adamjarling (approved):
Looking good.

PR #41, opened 2022-05-26
Style modal preview items.

mathewjordan:
## What does this do?

This adds some basic styling to the Preview component on the search filter modal screen. 

- Some slight reworking of the Figure was completed to allow for horizontal orientation. This is completed by passing an optional `orientation` prop to figure and rendering that to the DOM as a data attribute on the `figure`.
- Items in the preview are also now clickable

image

- Add baseline Preview styling.
- Add figcaption, orientation data attr.
- Style figures within preview.
- Add accession if title is missing.
- Add tests and mock search shape.


adamjarling (approved):
Lookin sharp

PR #42, opened 2022-05-26
Drive UI display by a IIIF manifest

adamjarling:
## What does this PR do?

In essence, it renders out UI based on data from a Work's IIIF Manifest rather than our API data.  

Since most of our manifests currently referenced in the API Work response contain little to no metadata, this PR implements a helper function which constructs a bare-bones IIIF Presentation 3 manifest.  This manifest object is then delivered to the Work page (at build time), maintaining the static page delivered to end users.

Next, this PR utilizes Nectar to easily display IIIF data on the page.  

image

### DIFF
--- a/lib/iiif/manifest-helpers.ts
+++ b/lib/iiif/manifest-helpers.ts
@@ -1,0 +1,71 @@
+import { Manifest } from "@iiif/presentation-3";
+import { WorkShape } from "@/types/components/works";
+import { getAPIData } from "@/lib/dc-api";
+import { manifestPres3Template } from "@/lib/iiif/manifest-pres3-template";
+
+interface MetadataInput {
+  label: string;
+  value: string[] | string;
+}
+export const buildMetadataValues = (metadata: MetadataInput[]) => {
+  return metadata.map(({ label, value }) => ({
+    label: {
+      none: [label],
+    },
+    value: {
+      none: Array.isArray(value) ? [...value] : [value],
+    },
+  }));
+};
+
+export const buildPres3Manifest = async (
+  work: WorkShape
+): Promise<Manifest | undefined> => {
+  const manifest = { ...manifestPres3Template };
+  try {
+    manifest.id = work.iiif_manifest;
+    manifest.label.none = [work.title];
+    manifest.summary &&
+      (manifest.summary.none = work.descriptions.map(
+        (description) => description
+      ));
+    manifest.metadata = buildMetadataValues([
+      {
+        label: "Contributor",
+        value: [...work.contributor_labels],
+      },
+      {
+        label: "Date",
+        value: [...work.dates_created],
+      },
+      {
+        label: "Identifier",
+        value: [...work.identifiers],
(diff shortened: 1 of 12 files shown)


lib/iiif/manifest-helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+        label: "Subjects",
+        value: [...work.subject_labels],
+      },
+      {
+        label: "Accession Number",
+        value: work.accession_number,
+      },
+      {
+        label: "Terms of Use",
+        value: work.terms_of_use,
```
mathewjordan, line 59: This is temporary, so no big deal, but I think long term, we'll want this metadata value as an entry in the `requiredStatement` value array and not here. More fitting place for it IIIF wise.

lib/iiif/manifest-helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+    ]);
+  } catch (err) {
+    throw new Error("Error building manifest locally");
+  }
+  return manifest;
+};
+
+export const getManifest = async (id: string) => {
+  const manifest = await getAPIData<Manifest>({ url: id });
```
mathewjordan, line 69: Simple enough!

types/declarations.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Declare any modules here which don't have
+ * types exported with their package.  Without
+ * this, TypeScript complains.
+ */
+
+declare module "@samvera/nectar-iiif";
```
mathewjordan, line 7: This is nice to have for packages, especially ones out of control. We should probably get on creating types for Nectar.

mathewjordan (approved):
Looks great.

mathewjordan (commented):
It looks like we have some build errors related to Nectar. I think this is related to the `dompurify` package that Nectar uses for sanitizing markup. Apparently that package doesn't play well with node environments and we may want to use https://www.npmjs.com/package/isomorphic-dompurify instead. At any rate, maybe we can pair quickly tomorrow to resolve? Everything else looks good. Seems to be only on build.

lib/iiif/manifest-helpers.ts
adamjarling: Sure.  I was kind of just throwing random metadata into this for now to get some content length happening.  Sounds good on `requiredStatement`.

lib/iiif/manifest-helpers.ts
adamjarling: Yea, maybe there's a better way, but by passing a generic into `getApiData` (ie. `Manifest` here), we can keep `getAPIData`  more simple and put the burden of type checking on the consumers of the function?   Guess that's like TypeScript 101 finally sinking in:)

adamjarling:
@mathewjordan I'm going to merge this branch, as seems unrelated to the Nectar dependency build issue.  And then we make a separate PR for the Nectar upgrade how about?

PR #43, opened 2022-05-27
Handle network errors so the NextJS build doesn't break

adamjarling:
## What does this do?

Handles network request errors so 
1. NextJS build doesn't break
2. The Work page shows (for now) and error message but still loads


mathewjordan (approved):
Looking solid. This will be nice to have no matter the state of the API.

PR #45, opened 2022-06-03
Add placeholder dialog windows for Work action buttons

adamjarling:
## What does this do?
Adds very basic dialog stubs for the Work action buttons.

image

image

It pulls (and replaces) some styles already defined for the `Filter` modal component, for more generic use case.   Eventually we might want to further consolidate the two, but for now this gets us going.


mathewjordan (approved):
Ready to go. Love what you did to use a single component. I wonder if next step is thinking about this from a design system standpoint?

PR #48, opened 2022-07-11
Create fixed search bar on scroll.

mathewjordan:
## What does this do?
This adds functionality to set the `Search` bar fixed to the top of the window onScoll to a certain point. Functionality for this is implemented using a new hook `useElementPosition` that watches the position in relation to the top absolute top of the window. If this number is greater than or equal to  `0`, a boolean value is assigned tot he `searchFixed` property in the search context. We now can use this value across our various components that may need to know this.

In the process of creating this functionality, some basic design choices were implemented to make the **Filter** button on the search page center itself and remain fixed. While testing this, it was discovered that the raxdix/popover (as a 0.1.7 release candidate) component was having issues positioning the content pane (similar to the issue in Meadow). To resolve this, I swapped out popover for dropdown-menu and used the most recent stable version, installing it with the `--legacy-peer-deps`

**Default View**
image

**On Scoll**
image

## How should we review?

- The Search page is the best place to look. Play around in there and scroll through some longer lists of items. Toggle on a few facets and try it too. Note that there is some jitteryness if the Filter userfacets are expanded and a scoll occurs. I think this might just need to be a future issue?
- Search will fixed site wide. Take it for a spin on a work page as well.

### DIFF
--- a/components/Header/Primary.tsx
+++ b/components/Header/Primary.tsx
@@ -1,33 +1,56 @@
 import { Primary, PrimaryInner } from "@/components/Header/Header.styled";
+import { useEffect, useRef, useState } from "react";
 import Container from "@/components/Shared/Container";
 import Heading from "@/components/Heading/Heading";
 import Link from "next/link";
 import Nav from "@/components/Nav/Nav";
 import Search from "@/components/Search/Search";
-import { useState } from "react";
+import useElementPosition from "@/hooks/useElementPosition";
+import { useSearchState } from "@/context/search-context";
 
 const HeaderPrimary: React.FC = () => {
-  const [searchActive, setSearchActive] = useState(false);
+  const [searchActive, setSearchActive] = useState<boolean>(false);
+
+  const { searchDispatch, searchState } = useSearchState();
+  const { searchFixed } = searchState;
+
+  const primaryRef = useRef<HTMLDivElement>(null);
+  const scrollPosition = useElementPosition(primaryRef);
+
+  useEffect(
+    () =>
+      searchDispatch({
+        searchFixed: scrollPosition >= 0,
+        type: "updateSearchFixed",
+      }),
+    [searchDispatch, scrollPosition]
+  );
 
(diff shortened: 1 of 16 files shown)


components/Header/Primary.tsx
```diff
@@ -1,33 +1,56 @@
...
-          <Nav>
-            <Link href="/collections">Browse Collections</Link>
-          </Nav>
-        </PrimaryInner>
-      </Container>
-    </Primary>
+    <>
+      <Primary
+        data-search-active={searchActive}
+        data-search-fixed={searchFixed}
```
adamjarling, line 37: In general, assume there's not a "React" way for `Primary` to understand a `prop`?   No big deal, as long as we can communicate the value, just curious.

hooks/useElementPosition.ts
```diff
@@ -0,0 +1,23 @@
...
+    window.addEventListener("scroll", updatePosition);
+    updatePosition();
+
+    return () => window.removeEventListener("scroll", updatePosition);
+  }, [element]);
+
+  return position;
+};
+
+export default useElementPosition;
```
adamjarling, line 23: Also curious how heavy this is in terms of browser memory?   Think putting some kind of debounce on it to reduce the computations would 1.) Impact UX, or 2.) event make a difference?    I like the fact that we're only calling it once in the app tho, perhaps that's optimization enough?

adamjarling (approved):
Looking good.

components/Header/Primary.tsx
mathewjordan: Thought about pass all props but then got into some funky typescript issues.

hooks/useElementPosition.ts
mathewjordan: I looked into using debounce but ran into some issues using it within the hook. Can't remember the exactly error, but something about hooks not being allowed within a functional component. May have been just me misunderstanding things. Maybe we should keep an eye on it and implement if we see any performance issues?

PR #50, opened 2022-07-13
Add custom server to support Auth

adamjarling:
Add custom server to the app which will run a `devbox` url locally so we can hit NU SSO auth stuff. 

Partial PR.  Fire the app up the same way: `npm run dev`

### DIFF
--- a/pages/api/auth/callback.ts
+++ b/pages/api/auth/callback.ts
@@ -1,0 +1,17 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  console.log("CALLBACK req.body", req);
+
+  switch (req.method) {
+    case "GET": {
+      // Construct callback URL
+      console.log("CALLBACK req", req);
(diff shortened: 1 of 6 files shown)


pages/api/auth/callback.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  console.log("CALLBACK req.body", req);
+
+  switch (req.method) {
+    case "GET": {
+      // Construct callback URL
+      console.log("CALLBACK req", req);
+
+      res.send("Hey");
```
mathewjordan, line 11: Still need this?

mathewjordan (approved):
Loads up great for me.

pages/api/auth/callback.ts
adamjarling: This will only be a partial PR @mathewjordan .  There's a lot more coming, but wanted to get this, kinda isolated piece in.  We'll clean it up in the next PR.

PR #55, opened 2022-07-18
Add footer component

adamjarling:
Pulls in a `Footer` component from https://github.com/nulib/design-system.   The current implementation is an example of how it could be styled...

image


components/Footer/Footer.tsx
```diff
@@ -0,0 +1,15 @@
+import Container from "@/components/Shared/Container";
+import { FooterStyled } from "@/components/Footer/Footer.styled";
+import { Footer as NUFooter } from "@nulib/design-system";
+import React from "react";
+import colors from "@/styles/colors";
```
mathewjordan, line 5: Nice.

mathewjordan (approved):
🚀 Yeah, seems good for now. We can comb back over it a bit later on to adjust styling if needed.

PR #56, opened 2022-07-18
Add Download and Share and a Work Context

mathewjordan:
## What does this do?

This one is a doozy.

image

**It:**

1. Add a Work Context 
2. Add some generic styling structures for Work dialog boxes using an Aside (including a Card) and Content component
3. Updates the IIIF manifest helpers to include items thumbnails
4. Implements layout decisions and most of the functionality for the Download and Share modal
5. Adds a shared component for social icons
6. Adds a shared component for copy text functionality

**It does not:**

1. Include Embed code functionality on Download and Share modal
2. Include Social Icon opengraph stuff
3. Fully complete **Download** functionality for Download and Embed (should we use `@samvera/image-downloader`?)
4. Handle responsiveness fully
5. Introduce any lazy loading on in the Download and Embed section (this could be long for some works)

## How should we review?
 - Go to any work (preferably [one with multiple filesets](https://devbox.library.northwestern.edu:3000/works/8db92d1e-b4a5-40c9-bdd8-e78d2837e37d))
 - Click Download and Share


adamjarling (approved):
Everything works great and like the consolidation.  My only question would be, if we're never updating a `manifest` or `work`, could we just declare a `React Context` in `pages/works/[id].tsx`, instead of the entire `work-context-tsx` file?   No big deal, just might be a pattern we potentially don't need all of here.

PR #58, opened 2022-07-22
3003 collection page baseline

adamjarling:
## What does this do?
Creates a baseline placeholder for individual Collection page in our app.   Note all the data is stubbed and not coming from a real Collection record.

image

## How to test
- Go to a Work 
- Click on the Collection image to visit the page
- Bask in the beauty of the page


components/Shared/Card.tsx
```diff
@@ -1,15 +1,19 @@
 import Figure from "@/components/Figure/Figure";
+import Link from "next/link";
+import { LinkedAnchor } from "@/components/Shared/Card.styled";
 import React from "react";
 
 export interface CardProps {
   description?: string;
+  href?: string;
```
mathewjordan, line 8: I like this pass through.

components/Work/TopInfo.tsx
```diff
@@ -87,8 +87,9 @@ const WorkTopInfo: React.FC<TopInfoProps> = ({ manifest, work }) => {
       <div>
         <h2>Collection</h2>
         <Card
-          title={work.collection_title}
-          description="Collection description"
+          title={`Ima title`}
+          description="Foo bar oh my"
```
mathewjordan, line 91: This somehow was reverted back to the **Ima Title**

```suggestion
          title={work.collection_title}
          description="Collection description"
```

mathewjordan (changes requested):
Just one small change where things reverted. Looks good otherwise and I think its a good start.

components/Work/TopInfo.tsx
adamjarling: Good catch.  Changes now updated.

mathewjordan (approved):
🚀 Let's go.

davidschober:
Screen Shot 2022-08-02 at 2 43 02 PM


^^^ This part should be an overlay pinned to the bottom of every page on the site that a user can click "acknowledge" on . #2695

PR #60, opened 2022-07-28
Improve / Fix Work Facet filtering UX

adamjarling:
## What does this do?
- Maintains focus state on Facets text filter `input` element
- Adds a small `debounce` to the text filter
- Refactors the location of aggregations data fetching in the `Facet` component to avoid clunky re-renders
- Adds a quick loading spinner

image

## To Test
- Go to the Search screen and click the "Filter" button in top menu
- Click into a Facet display where you see all the values
- Text filter via the input box
- Verify if works with user selected facets.  Check some boxes.

### DIFF
--- a/components/Shared/Loader.styled.ts
+++ b/components/Shared/Loader.styled.ts
@@ -1,0 +1,21 @@
+import { keyframes, styled } from "@/stitches.config";
+
+/* eslint sort-keys: 0 */
+
+const rotation = keyframes({
+  "0%": { transform: "rotate(0deg)" },
+  "100%": { transform: "rotate(360deg)" },
+});
+
+const SpinLoader = styled("span", {
+  width: "48px",
+  height: "48px",
+  border: "5px solid $black10",
+  borderBottomColor: "transparent",
+  borderRadius: "50%",
+  display: "inline-block",
+  boxSizing: "border-box",
+  animation: `${rotation} 1s linear infinite`,
+});
+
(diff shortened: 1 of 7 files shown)


components/Shared/Loader.styled.ts
```diff
@@ -0,0 +1,21 @@
...
+const SpinLoader = styled("span", {
+  width: "48px",
+  height: "48px",
+  border: "5px solid $black10",
+  borderBottomColor: "transparent",
+  borderRadius: "50%",
+  display: "inline-block",
+  boxSizing: "border-box",
+  animation: `${rotation} 1s linear infinite`,
+});
```
mathewjordan, line 19: Nice.

mathewjordan (changes requested):
Overall, seems to work great. Looks like our query is case sensitive right now. That could be a separate issue altogether and something I just never noticed before.

"ber" and "Ber" provide different results.

adamjarling:
> Overall, seems to work great. Looks like our query is case sensitive right now. That could be a separate issue altogether and something I just never noticed before.
> 
> "ber" and "Ber" provide different results.

@mathewjordan I was going to try and fix this, but looks like the search term matching logic is in the API itself:  (https://zqwev40ued.execute-api.us-east-1.amazonaws.com/test) so maybe that should be a separate issue?

mathewjordan (approved):
Totally makes sense. Let's handle it elsewhere.

PR #61, opened 2022-07-29
Style Work page.

mathewjordan:
## What does this do?

This work brings in some styling decisions on the Work page that should influence the direction of Digital Collections. Along with it, colors and fonts are beginning to be incorporated more intently. Also added is the Northwestern SVG logo.

image

In general, decisions made are:

- Take advantage of whitespace/negative areas and let stuff breath
- Big bold titles using **Campton Bold** for Works, Collections, and Related links
- Utilize purple (below the header) more intently as an **action** color for links and buttons
- Tick up the font size for body content


components/Shared/DefinitionList.styled.ts
```diff
@@ -3,15 +3,17 @@ import { styled } from "@/stitches.config";
...
 
 const DefinitionListWrapper = styled("div", {
+  lineHeight: "1.47em",
+
   "& dt": {
-    fontSize: "0.75rem",
-    fontWeight: "bold",
-    paddingBottom: "0.5rem",
-    textTransform: "uppercase",
+    fontSize: "$3",
```
adamjarling, line 9: Great, I've been looking for these.

components/Work/TopInfo.styled.ts
```diff
@@ -5,26 +5,77 @@ import { styled } from "@/stitches.config";
...
+      color: "$black50",
+      fontFamily: "$sansLight",
+      lineHeight: "1.47em",
+    },
+  },
+
+  [`> div`]: {
+    display: "grid",
+    gap: "$7",
+    gridTemplateColumns: "618fr 382fr",
```
adamjarling, line 62: Wow, never seen `fr` values this big!   Can you explain a little why?

components/Work/ViewerWrapper.tsx
```diff
@@ -2,21 +2,56 @@ import React from "react";
...
+}> = dynamic(() => import("@samvera/clover-iiif"), {
+  ssr: false,
+});
 
 interface WrapperProps {
   manifestId: string;
 }
 
 const WorkViewerWrapper: React.FC<WrapperProps> = ({ manifestId }) => {
+  const customTheme = {
```
adamjarling, line 21: Nice

styles/global.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+  ],
+};
+
+const globalStyles = globalCss({
+  ...defaults,
+  ...fonts,
+});
+
+export default globalStyles;
```
adamjarling, line 84: Wow, this is cool.  So this is the Stitches way of injecting globals eh?   Seems clean to me.

adamjarling (approved):
This is fantastic work.  Looks great!

styles/global.ts
mathewjordan: Yup, just discovered this. Its not well documented and took some trial and error. Seems like a good idea for a medium article.

components/Work/TopInfo.styled.ts
mathewjordan: It's a golden ratio calculation, not absolutely necessary.

PR #63, opened 2022-08-02
Create Expand component.

mathewjordan:
## What does this do?

This adds an `<Expand>` component to our Shared components library that wraps child content and includes a Button for expanding content vertically. I opted for creating this as a shared component for reuse elsewhere, likely on the Collections level pages.

image

onClick of the `ExpandButton`, the **maxHeight** will adjust to the actual height of the `ExpandContent` element. A CSS transition adds a smooth push down of content that is rendered below the `<Expand>`. Defaults are set for both the **buttonText** ("Show More") and **initialHeight** (500) values for smoother implementation.

### DIFF
--- a/components/Figure/Figure.styled.ts
+++ b/components/Figure/Figure.styled.ts
@@ -41,6 +41,8 @@ const FigureStyled = styled("figure", {
   variants: {
     isPromoted: {
       true: {
+        [`& ${Image}`]: { maxHeight: "200px" },
+
         [`& ${Title}`]: {
           fontSize: "$5",
           fontFamily: "$displayBold",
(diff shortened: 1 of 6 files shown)


adamjarling (approved):
This looks and works great.  Really nice work.  Just curious if it needs a "Less" button or anything to close up the expanded area?   Or just one and done for the user interacting with the element?

mathewjordan:
> This looks and works great. Really nice work. Just curious if it needs a "Less" button or anything to close up the expanded area? Or just one and done for the user interacting with the element?

Definitely considered this but felt like that was probably functionality that wouldn't be used. Looked at a couple of examples around the web where this type of **Expand** or **Show More** functionality existed and mostly places seem to shy away from any action to collapse except for accordion or type functionality. Going without it feels cleaner but we can always have it be a true on/off toggle at some point if its requested.

PR #64, opened 2022-08-04
3008 homepage design

mathewjordan:
- Add boolean hero variant on header.
- Update sizes.
- Rectify sizing.


mathewjordan:
See https://github.com/nulib/dc-nextjs/pull/66

PR #65, opened 2022-08-05
3005 collection page aggs

adamjarling:
Potentially partial PR to get some real (as of current state of API) data powering the Collection page.

@mathewjordan Figure we can chat through where to take this, this afternoon?    The "Metadata" tab is active by default and will initially show a listing of "Subjects" pulled in via aggs on a Collection's Works.   I imagine next step is to extend "Subjects" to "Genre" or other metadata items, then pull those in dynamically?  

The "Explore" just has placeholder Bloom sliders in 'em.   "Collection Organization" tab does nothing for now.

image


adamjarling:
There are no tests yet, intentional, until I get some confirmation I'm on the right track :)

mathewjordan (approved):
Let's shoot this through and iterate. :shipit:

PR #66, opened 2022-08-05
Add Hero component using swiper.js

mathewjordan:
## What does this do?

image

This integrates swiper.js and and Hero component to the homepage. This Hero component current renders content from a IIIF collection and someday could be a standalone IIIF component in samvera-labs.

Also introduced are custom sizes/spaces/fontSizes for stitches.


adamjarling (approved):
Looking really good.

PR #68, opened 2022-08-15
Create homepage interstitial

mathewjordan:
## What does this do?

This adds a Homepage interstitial for introducing and describing in basic terms. This section will be render below the homepage hero. Text content included here is borrowed from JSTOR and should not be used in production and should probably be written by the copywriting talent in RDC.

Additionally, this PR tackles the outstanding issue regarding sizes/spaces in stitches by defining sizes/spaces in a naming pattern prefixed with `gr`. This was necessary as stitches default sizing tokens were not playing well with Next.

image


adamjarling (approved):
Sharp.  The homepage is really coming together.

PR #69, opened 2022-08-16
3004 update new api endpoints

adamjarling:
This PR updates TypeScript Types and the UI in general to pull from the new API endpoints and match (mostly, probably some bugs still lurking) the UI to conform to the new shape.


mathewjordan (commented):
Two things I'm noticing:

Maybe it's on my side only... I did the update to miscellany. Something else I need to do?

All works seem to give the "Aw. shucks..." message 
image

And all search results are missing their title/work type
image

mathewjordan (approved):
Looking good for the most part. Gonna approve since I'm not sure if this is another expected issue. Just going to note in case we need to spin off a new one and think of this work as incremental.

The Work Type facets (aligned right) do not seem to work right now for filtering things.
image


And under the the Filter modal, the Work Type tab does not display options.

adamjarling:
Pushing work to a new preview branch

PR #73, opened 2022-08-17
Fix work types filtering

adamjarling:
This PR fixes the Work Type filters which were broken after new API implementation

image


mathewjordan (approved):
Good now! 🚀

PR #74, opened 2022-08-17
Create dismissable global 'harmful content' message.

mathewjordan:
## What does this do?

This work adds a dismissable message fixed to the bottom of all pages. Data related to this message for `status` and `timestamp` are kept in browser session storage. The current interval for dismissing a message is `86400` seconds (24 hours). Once that that time has elapsed, the message will display once again for the user of that browser.
 
image

### DIFF
--- a/hooks/useEventCallback.ts
+++ b/hooks/useEventCallback.ts
@@ -1,0 +1,17 @@
+import { useCallback, useRef } from "react";
+
+import useIsomorphicLayoutEffect from "@/hooks/useIsomorphicLayoutEffect";
+
+export default function useEventCallback<Args extends unknown[], R>(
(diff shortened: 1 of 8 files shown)


hooks/useEventCallback.ts
```diff
@@ -0,0 +1,17 @@
+import { useCallback, useRef } from "react";
+
+import useIsomorphicLayoutEffect from "@/hooks/useIsomorphicLayoutEffect";
```
adamjarling, line 3: Quite a mouthful.  I'm scared of this hook:)

adamjarling (approved):
Looks like a lot of work here.  Maybe you can run me through it Friday?

PR #79, opened 2022-08-24
Update bloom, dynamic import for swiper integration.

mathewjordan:
## What does this do?

This updates Bloom to a version supporting baseline responsiveness (using Swiper) that works with Next.js. I also switched it to a dynamic import --- required by the swiper dependency.

image


components/Shared/RelatedItems.tsx
```diff
@@ -1,13 +1,19 @@
...
   collections?: string[];
   title: string;
 }
 
+export const BloomIIIF: React.ComponentType<{
+  collectionId: string;
+}> = dynamic(() => import("@samvera/bloom-iiif"), {
+  ssr: false,
+});
+
```
adamjarling, line 16: I keep forgetting about this trick.  Nice.

adamjarling (approved):

PR #80, opened 2022-08-30
UI cleanup

adamjarling:
## What does this do?

- Fixes the harmful content display issue in UI
- Experiments with `framer-motion` as a way to handle page transitions.

@mathewjordan Take a look at this.  Not sure we need `framer-motion`, but just messed around with it briefly to do some heavy lifting if we wanted to dig in more.  Curious on your thoughts.  Not refined, but an ideas I suppose.

### DIFF
--- a/components/Header/Super.tsx
+++ b/components/Header/Super.tsx
@@ -25,13 +25,17 @@ const nav = [
 ];
 
 export default function HeaderSuper() {
+  const [isLoaded, setIsLoaded] = React.useState(false);
+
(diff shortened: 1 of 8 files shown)


mathewjordan (approved):
I actually really like this approach. Never really thought to try it this way but seems like a tidy way to handle things.

adamjarling:
@mathewjordan fyi I'm going to leave these 2 commits in place, as they seem not logically grouped in case we wanted to remove `framer-motion` at some point.

PR #81, opened 2022-09-01
2744 pagination setup

adamjarling:
## What does this do?
- This handles pagination and makes use of the `pagination` object returned from a `/search` API request.   
- It's a first approach to making the Search Page component aware of pagination, and how and when to fetch paginated data.  Future iterations can build from this initial pattern.
- Pagination is synced to browser url `history` API
- Some base level styling

## What this does *not* do
- It does not update the browser URL when new facets are updated.  It should, but outside the scope of this ticket perhaps.

image

## To test
1. Go to the search page and interact with pagination links
2. Try doing a new search, or adding some facets and observe the behavior.

### DIFF
--- a/components/Search/Pagination.styled.ts
+++ b/components/Search/Pagination.styled.ts
@@ -1,0 +1,93 @@
+import { styled } from "@/stitches.config";
+
+const linkActiveVariant = {
+  active: {
+    false: {
+      visibility: "hidden",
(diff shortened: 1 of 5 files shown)


kdid (approved):

mathewjordan:
Just noting that I really like where this ended visually.

PR #82, opened 2022-09-01
An idea for how to eventually handle images loading

adamjarling:
## What does this do?

This can be considered a "partial PR", which uses a lightweight, customizable enough package which we might use to serve as placeholder content for loading images, documents, players, whatever.

We still need the dimensions of the Work's representative image, to make this work proper with a masonry grid layout.

image


components/Shared/LoadingBox.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <ContentLoader
+      speed={2}
+      width={400}
+      height={460}
+      viewBox="0 0 400 460"
+      backgroundColor="#f3f3f3"
+      foregroundColor="#ecebeb"
+      {...props}
+    />
```
mathewjordan, line 14:
```suggestion
    <ContentLoader
      speed={2}
      width={400}
      height={460}
      viewBox="0 0 400 460"
      backgroundColor="#f3f3f3"
      foregroundColor="#ecebeb"
      {...props}
    >
      <rect x="0" y="0" width="200" height="200" />
      <rect x="0" y="218" width="180" height="8" />
      <rect x="0" y="230" width="120" height="8" />
    </ContentLoader>
```

mathewjordan (changes requested):
Similar to what I was thinking. A few notes: 
- Remove `public/images/book_placeholder.png` -- seems like we don't need this
- With react-content-loader we can add child `<svg>`s. I've added a suggestion for that.

mathewjordan (approved):

PR #83, opened 2022-09-12
Support deep linking of user selected facets

adamjarling:
## What does this do?

- Moves tracking/handling user selected facets out of the `SearchProvider` component.
- Current user facets are now read and updated by individual components, via `Next/Link`'s routing object.
- Tests updated to reflect a change in the `SearchProvider` component and supports mocking Next's Router package
- In the Search page component, we bypass the `useFetchApiData` hook and grab the information we want directly from the component.  There were a lot of timing errors relying upon the convenience hook (for some reason), and after trying a bunch of different approaches, this seemed to simplify things and produce consistent results.

image

## To Test
1. Go to the /search page
2. Try a combination of searching for terms, applying facets, removing facets, etc.
3. New search terms should wipe out previous facet selections.   
4. New facet selections should retain the search query, and narrow down the results
5. Try doing hard refreshes in the browser and notice we can save complex Search page filtering now

### DIFF
--- a/lib/queries/builder.ts
+++ b/lib/queries/builder.ts
@@ -1,7 +1,7 @@
 import { buildSearchPart, querySearchTemplate } from "@/lib/queries/search";
 import { ApiSearchRequest } from "@/types/api/request";
 import { FacetsInstance } from "@/types/components/facets";
-import { UserFacets } from "@/types/context/search-context";
+import { UrlFacets } from "@/types/context/filter-context";
 import { buildAggs } from "@/lib/queries/aggs";
 import { buildFacetPart } from "@/lib/queries/facet";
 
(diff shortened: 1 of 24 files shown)


types/context/filter-context.ts
```diff
@@ -1,9 +1,13 @@
 import { FacetsInstance } from "@/types/components/facets";
 export interface FilterContextStore {
   recentFacet?: FacetsInstance;
-  userFacetsUnsubmitted: UserFacetsUnsubmitted;
+  userFacetsUnsubmitted: UrlFacets;
```
mathewjordan, line 4: We may want to rename `userFacetsUnsubmitted` but not a big issue at this point.

lib/queries/builder.ts
```diff
@@ -10,7 +10,7 @@ type BuildQueryProps = {
   aggsFilterValue?: string;
   size?: number;
   term: string;
-  userFacets: UserFacets;
+  userFacets: UrlFacets;
```
mathewjordan, line 13: rename this as well?

mathewjordan (approved):
This looks good. Thanks for the walk through. Just a few things we may want to rename but otherwise no issues that I can spot.

types/context/filter-context.ts
adamjarling: Yea, thought about this one, and at least to me the name makes sense.  It's user facet selections, they haven't been applied yet, so they're not in the URL quite yet?

PR #84, opened 2022-09-19
3152 cypress search page refactor idea

adamjarling:
## What does this do?

1. Installs Cypress as end to end testing framework.  Some initial setup to test the `/homepage`, `/search`, `/item` routes.  Mainly just an experiment and NOT fully fleshed out.   See https://github.com/nulib/repodev_planning_and_docs/issues/3152 for more detailed notes.

2. Bug fixes to url-based search and facets query parameter approach


mathewjordan (approved):
Big exploration here. I like seeing the tests in action, especially on the search modal. Even given the complexities you outlined with SSG and e2e, I still think there is room for us to eventually have this in place in our workflow.

PR #85, opened 2022-09-27
Create Browse Collections page.

mathewjordan:
## What does this do?

This work adds a collection list page a `/collections` and renders out each item in a vertical list. The results are filterable for a simple substring lookup on the collection title. Some small workarounds are baked into this to build the full collection list by jump through `next_url` on each page request until it is undefined.  I did toy with thee idea of bringing Bloom in here for some sort of Collection preview functionality. Ultimately I found it a bit cumbersome. I'm open to coming back to it but feel we might want to chat through that a bit.

Notes:

- Thumbnail is rendered as a gray placeholder if it `representative_image.work_id` does not exist on the collection.
- A `ReadMore` component was created to expand out text strings for collection descriptions

image


adamjarling (approved):
Looks great

PR #87, opened 2022-09-30
Update collection as IIIF endpoint.

mathewjordan:


mathewjordan:
`DC_API_SEARCH_IIIF_URL` Still exists in **main.tf** and **variables.tf**

PR #89, opened 2022-09-30
Update Search figure src to thumbnail endpoint.

mathewjordan:
## What does this do?

This change swaps out the `src` prop value to use the /`thumbnail` endpoint for search results on works.

image


kdid (approved):

mathewjordan:
Thanks @kdid

PR #94, opened 2022-10-10
Fix JSON LD Schema.org validation for type Collection

adamjarling:
Remove `thumbnail` property from JSON LD Schema Collection @type, so it validates.


mathewjordan (approved):
👍

kdid (approved):

PR #95, opened 2022-10-11
Show Collection image on Work page

adamjarling:
## What does this do?

Update Work page to show Collection representative image instead of the Work image (current placeholder).

## How to test
1. Go to any Work/Item
2. Verify Collection image is different from the Work/Item image.  Or if a Collection image does not exist, it will default to a placeholder image

image


mathewjordan (approved):
Solid.

PR #96, opened 2022-10-11
Collection page cleanup prior to user testing

adamjarling:
Quick UI cleanup


mathewjordan (approved):
Awesome, so much better.

PR #98, opened 2022-10-12
Implement social sharing via OpenGraph

adamjarling:
## What does this do?
Implements the [OpenGraph](https://ogp.me/) protocol to generate data when sharing app links.

image

image

## How to test
1. Run the app in your AWS environment (or just inspect locally in browser console)
2. On each page, verify opengraph data exists in browser console
3. If in AWS environment, you can use this validation tool to view OpenGraph data: https://www.opengraph.xyz/

### DIFF
--- a/components/layout.tsx
+++ b/components/layout.tsx
@@ -30,14 +30,6 @@ const Layout: React.FC<LayoutProps> = ({
       <Head>
         <link rel="icon" href="/favicon.ico" />
         <meta name="description" content="NUL Digital Collections v2" />
-        <meta
-          property="og:image"
-          content={`https://og-image.vercel.app/${encodeURI(
-            siteTitle
-          )}.png?theme=light&md=0&fontSize=75px&images=https%3A%2F%2Fassets.zeit.co%2Fimage%2Fupload%2Ffront%2Fassets%2Fdesign%2Fnextjs-black-logo.svg`}
-        />
-        <meta name="og:title" content={siteTitle} />
-        <meta name="twitter:card" content="summary_large_image" />
         <title>{title}</title>
       </Head>
       <Header isHero={header === "hero"} />
--- a/lib/json-ld.ts
+++ b/lib/json-ld.ts
@@ -1,9 +1,9 @@
 import { CollectionShape } from "@/types/components/collections";
+import { PRODUCTION_URL } from "@/lib/constants/endpoints";
 import { WorkShape } from "@/types/components/works";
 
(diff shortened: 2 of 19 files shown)


components/layout.tsx
```diff
@@ -30,14 +30,6 @@ const Layout: React.FC<LayoutProps> = ({
       <Head>
         <link rel="icon" href="/favicon.ico" />
         <meta name="description" content="NUL Digital Collections v2" />
-        <meta
```
mathewjordan, line 33: Oh, good. Glad to see that go.

lib/json-ld.ts
```diff
@@ -1,9 +1,9 @@
 import { CollectionShape } from "@/types/components/collections";
+import { PRODUCTION_URL } from "./constants/endpoints";
```
mathewjordan, line 2:
```suggestion
import { PRODUCTION_URL } from "@/lib/constants/endpoints";
```

lib/open-graph.ts
```diff
@@ -0,0 +1,55 @@
...
+  const ogTitle = work?.title ? work.title : work?.accession_number;
+  const isPublic = isPublicWork(work?.visibility);
+  const imageUrl =
+    isPublic && isImageType(work?.work_type)
+      ? `${work?.representative_file_set.url}/full/1200,630/0/default.jpg`
+      : work?.thumbnail;
+
+  let ogDescription = "";
+  if (work?.description && work.description.length > 0) {
+    ogDescription = `${work.description[0]} - `;
```
mathewjordan, line 34: Is it possible for us to want to deliver all values of a description? Maybe something like this in case?

```suggestion
    ogDescription = `${work.description.join(" ")} - `;
```

lib/open-graph.ts
```diff
@@ -0,0 +1,55 @@
+import { ObjectLiteral, OpenGraphData } from "@/types/index";
+import { isImageType, isPublicWork } from "./work-helpers";
```
mathewjordan, line 2:
```suggestion
import { isImageType, isPublicWork } from "@/lib/work-helpers";
```

mathewjordan (changes requested):
Just three small things... thoughts on `.join()` for the `work.description`?

lib/open-graph.ts
adamjarling: Sure thing

mathewjordan (approved):
Looks great.

PR #99, opened 2022-10-13
Add Embed HTML copyable snippet to Work Download and Embed modal

adamjarling:
## What does this do?
Replicates the HTML Embed functionality from DC v1.

image

## How to test
1. Open any work in DC v2 and click the "Download and share" button.
2. Click "Embed HTML"
3. Verify the width and color selections change in the UI.
4. Verify the copy functionality works

### DIFF
--- a/components/Work/ActionsDialog/DownloadAndShare.tsx
+++ b/components/Work/ActionsDialog/DownloadAndShare.tsx
@@ -4,19 +4,22 @@ import {
 } from "@/components/Work/ActionsDialog/ActionsDialog.styled";
 import { Canvas, IIIFExternalWebResource } from "@iiif/presentation-3";
 import {
+  EmbedHTML,
+  EmbedHTMLActionRow,
   EmbedViewer,
   ItemActions,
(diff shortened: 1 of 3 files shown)


components/Work/ActionsDialog/DownloadAndShare.tsx
```diff
@@ -67,31 +70,102 @@ const DownloadAndShare: React.FC = () => {
...
+    },
+    {
+      label: "900px - 25%",
+      value: 900,
+    },
+    {
+      label: "450px - 12.5%",
+      value: 450,
+    },
+  ];
```
mathewjordan, line 109: Maybe we'll want to move this to a constants library at some point?

mathewjordan (approved):
This is super slick.

components/Work/ActionsDialog/DownloadAndShare.tsx
adamjarling: Sure, I guess if we use them more than once?

PR #100, opened 2022-10-13
Collection item styling.

mathewjordan:
## What does this do?

The work overhauls the Collection page to match more closely the styling of other parts of the site with a general simplification of design approach.

image

Covers:

- Integration of the Hero component for a collection item
- Updates to Hero component for more flexible usage
- Refinements to tabbing (closely matching Clover navigator tabs)
- Updates to the HeroCollection data model to allow for buttons using the `seeAlso` property
- Updates to "Search Collection" button to use `next/link`
- Moving description to the Explore/About tab.
- Making Explore/About the default tab.

image

Does not cover:
 - Data fetching for Related Items of a Collection
 - Extrapolating on the Subjects tab (anything more to do here?)
 - Collection Organization


lib/constants/endpoints.ts
```diff
@@ -1,5 +1,8 @@
 const DCAPI_ENDPOINT = process.env.NEXT_PUBLIC_DCAPI_ENDPOINT;
 const DC_API_SEARCH_URL = `${DCAPI_ENDPOINT}/search`;
+const DC_URL = process.env.DC_URL
```
adamjarling, line 3: Seems like we're both using `DC_URL` lately.   Wonder if we should adjust to title `NEXT_PUBLIC_DC_URL` so we can grab it from client and server side?   @mbklein @kdid

adamjarling (approved):
Way, way, way, way better!

lib/constants/endpoints.ts
kdid: Yes, I can add a new var - I'll write an issue

PR #102, opened 2022-10-19
Initiate Nectar metadata w/ links.

mathewjordan:
## What does this do?

The primary work here brings in functionality that  allows`next/link` to wrap prescribed metadata values.

To do so, this:

- incorporates updates from nectar that allow customization of Metadata output, this is source from the **FACETS_WORK_LINK** constant in `
@/lib/constants/works`.
- breaks out WorkMetadata out as a seperate component that is rendered in TopInfo
- and points the Work (items) at an actual Presentation 3.0 manifest provided from the `?as=iiif` endpoint. In the process of this, a lot of temporary functionality was removed that generated a temporary working presentation 3.0 manifest within Next.
- Updates global anchor styling

Some notes: Subject does not properly link. I believe we should make an update to the IIIF Manifest themselves to not render Subject values with their roles such as `(Topical)`.

image


lib/dc-api.ts
```diff
@@ -23,4 +23,19 @@ async function getAPIData<R>(obj: GetApiParams): Promise<R | undefined> {
...
+    });
+    const data = (await response.json()) as Promise<R>;
+    return data;
+  } catch (err) {
+    if (err instanceof Error) {
+      throw new Error(err.message);
+    }
+  }
+}
+export { getAPIData, getIIIFResource };
```
adamjarling, line 41: Think `getAPIData` would work for this as well, but if you wanted to be more specific this added function works too.

adamjarling (approved):
Looking good

PR #104, opened 2022-10-27
Add support for private or institutional works display on Work page

adamjarling:
## What does this do
On the Work page, handles scenarios for Work's which require authentication
- Adds a blurred image component, using a Work's `thumbnail` as the source.
- Adds an Announcement message indicating why there is a blurred image.
- Adds `visibility` facet, so we can easily find "Institution" and "Private" Works
- Updates the Northwestern favicon in browser

image

## How to test
1. Go to search page
2. Filter by "Rights and Usage" > "Visibility"
3. Select an "Institution" Work
4. Verify blurred image (only will appear if a `thumbnail` value is present, which we think will eventually get populated for all Works?)
5. Click "sign in" link in the announcement, and you'll scroll to top of the viewport

## Other notes
the `<BlurredBgImage />` component is somewhat flexible and we may be able to use this in other places.  Here's the component's Props types:

```
interface Props {
  bgColor?: string;
  bgImageUrl: string;
  height?: string;
  width?: string;
}
```


adamjarling:
@mathewjordan I'm more than happy to rebase this off whatever you having coming down the pipe.  Your call, guessing we did the same thing, or similar, to fix up the DC API v2 calls.

components/Shared/BlurredBgImage.tsx
```diff
@@ -0,0 +1,37 @@
...
+  bgImageUrl,
+  height,
+  width,
+  ...restProps
+}) => {
+  return (
+    <div
+      style={{
+        backgroundColor: bgColor,
+      }}
```
mathewjordan, line 22:
```suggestion
        backgroundColor: bgColor,
        overflow: "hidden",
      }}
```

mathewjordan: The filter blur on the image will bleed beyond the bounds of the wrapping div. adding overflow should solve that issue.

components/Shared/BlurredBgImage.tsx
```diff
@@ -0,0 +1,37 @@
...
+      data-testid="blurred-bg-image"
+      {...restProps}
+    >
+      <BlurredBgImageStyled
+        css={{
+          backgroundImage: `url(${bgImageUrl})`,
```
mathewjordan, line 28: Since we are blurring, we could get away sending a smaller thumbnail. Not necessary but saves some data.
https://dcapi.rdc-staging.library.northwestern.edu/api/v2/works/6c2e4d68-c55e-4f3b-91ad-ae37bb63ba39/thumbnail?size=20

```suggestion
          backgroundImage: `url(${bgImageUrl}?size=20)`,
```

components/Work/RestrictedDisplay.tsx
```diff
@@ -0,0 +1,47 @@
...
+            sign in
+          </a>{" "}
+          or email{" "}
+          <a href="mailto:repository@northwestern.edu">
+            repository@northwestern.edu
+          </a>{" "}
+          to request access to this work.
+        </p>
+      </Announcement>
+    </div>
```
mathewjordan, line 43: Should we consider rendering a small thumbnail in here?

mathewjordan (approved):
I like it. Made a few suggestions but I don't think they should block the progress.

components/Shared/BlurredBgImage.tsx
adamjarling: Sure.  I didn't even know there were query params possible on this endpoint.

components/Work/RestrictedDisplay.tsx
adamjarling: In addition to the image above it?   Not sure I get it w/o seeing it, but open to whatever.

PR #105, opened 2022-10-28
Modal filter enhancements.

mathewjordan:
## What does this do?

This brings in some styling refinements for the Filter modal.

- Adds chevron on Group tabs
- Refines styling modal wide for font treatments
- Standardizing font sizes
- Blue Indicator of active facet
- Subject as the default Facet selected
- Language update for _Preview Results_
- Responsiveness on modal width/height

image


adamjarling (approved):
Waaaay better.  Looks and feels both easy to use and understand.  Excellent.

PR #106, opened 2022-10-31
Update logout to use API logout route

kdid:
### Summary

- Use DC API V2's `/auth/logout` route which will take you to the NUSSO logout form

### How to Test

- New browser/cookies get rid of any current DC session
- Run DC next in your dev environment (*with https and either using devbox locally or your dev prefix in remote dev)
- Login
- Click logout and see that you are presented with the NUSSO logout screen. 
- Logout

** I will apply the Terraform changes after this is merged.


adamjarling (approved):
Looks good to me.

PR #108, opened 2022-11-02
Build out Collections page Explore tab data

adamjarling:
## What does this do?
- Grabs the 10 most popular Work subject metadata values for a NU Collection, then selects 3 random "popular" subjects.  The Explore (About) tab of a Collection page will then query the API to get Collection Works for each subject.  It then feeds `Bloom` these IIIF collection uris to build out sliders.
- Minor style update to the "focus" state of an active content tab on a Collection page (from black to light grey).
- Working update for a Restricted Work on the Work page.

image

## To Test
Go to any Collection page, and if subject metadata exists for Works in that Collection, `Bloom` sliders will render.


mathewjordan (approved):
This looks good to me. Some cleanup to do in the bloom package but otherwise is solid.

PR #109, opened 2022-11-03
More Like This on Work.

mathewjordan:
## What does this do?

This work adds "More Like This" Bloom slider with a collection response from the /similar endpoint of a work. The actual integration of this was absolutely easy. In the process I also cleaned up the search query URIs for the genre and subject collections. Somewhere along the way, these began failing.

image

This work should be exclusive of any bug fixes or enhancements we make to Bloom itself.


adamjarling (approved):
Looks simple enough and nice on the cleanup.

PR #110, opened 2022-11-07
Add shared link route /shared, to the app

adamjarling:


kdid:
Hey @adamjarling - I'm not sure how i would test this on a work. Public/published works don't create shared links, they just spit out a link to DC's regular route?

adamjarling:
@kdid Hmm, well maybe there is no way to really test this code.  I guess it'd be that it loads a Public shared Work for now?   I think the rush of getting this in, is keeping our Types in sync for Typescript, so we don't have to double up work elsewhere.

kdid:
It should not display a public work through the shared route (and it did not). I'll approve, the code looks 👍 to me

PR #113, opened 2022-11-09
Collection Organization, and collection revamp

mathewjordan:
image

## What does this do?

This work incorporates Radix UI Accordion components (as an `<ExpandableList>`) to order a collection by Series. On expand each series will render a Bloom slider. 

- No item will be open by default.
- Only one list item may be open at any time. 
- The user can also choose collapse a list item that is open.

Also incorporated into this work is `<Facts>` component that will render a Big/Small text in combination. These values are currently hardcoded placeholders but should be mapped from aggs in a future issue.

A collection's `finding_aid_url` will be rendered to a Button in the hero if it is not null.

### DIFF
--- a/components/Heading/Heading.styled.ts
+++ b/components/Heading/Heading.styled.ts
@@ -29,6 +29,20 @@ const StyledHeading = styled("h2", {
       top: "-$gr4",
     },
   },
+
+  "&[data-level=h2]": {
+    fontFamily: "$displayBold",
+    fontSize: "$gr6",
+    fontWeight: "400",
+    marginBottom: "$gr5",
+  },
+
+  "&[data-level=h3]": {
+    fontFamily: "$displayBook",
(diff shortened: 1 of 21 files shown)


components/Heading/Heading.styled.ts
```diff
@@ -29,6 +29,20 @@ const StyledHeading = styled("h2", {
...
+    fontWeight: "400",
+    marginBottom: "$gr5",
+  },
+
+  "&[data-level=h3]": {
+    fontFamily: "$displayBook",
+    fontSize: "$gr6",
+    fontWeight: "400",
+    marginBottom: "$gr3",
+  },
```
adamjarling, line 45: Ok, I can see this approach and will need to update my stuff to match.

adamjarling (approved):
I like it.  Clean, to the point and gives nice starter insight into a Collection eh?

PR #114, opened 2022-11-14
3240 about contact

adamjarling:
## What does this do?
At the highest level, this adds content to the About and Contact pages.  Also included in the PR along for the ride are some style additions:

- `<PhotoFeature />` component (on About page)
- `<SectionTop />` component from NU global marketing
- `<HeroBasic />` component from NU global marketing's Hero.
- `<BlockStyled />` component for consistent block level spacing
- `<LinkStyled />` component as a beginning pattern for consistent `<a />` elements site wide.
- Updates to the `<Heading />` component to more closely align with NU global marketing header styles.

## How to test
Take a spin through the app and visit the About and Contact pages.


components/About/CollectionGrid.tsx
```diff
@@ -0,0 +1,27 @@
...
+      {items.map(({ callToAction, href, imgAlt, imgSrc, title }) => (
+        <PhotoFeature
+          key={imgSrc}
+          callToAction={callToAction}
+          href={href}
+          imgAlt={imgAlt}
+          imgSrc={imgSrc}
+          title={title}
+        />
+      ))}
```
mathewjordan, line 22: In the future, especially if we don't hardcode our grid here, maybe we source these images via IIIF somehow?

package.json
```diff
@@ -26,7 +26,7 @@
     "@radix-ui/react-radio-group": "^0.1.6-rc.20",
     "@radix-ui/react-select": "^1.0.0",
     "@radix-ui/react-tabs": "^0.1.6-rc.13",
-    "@samvera/bloom-iiif": "^0.3.3",
+    "@samvera/bloom-iiif": "^0.3.4",
```
mathewjordan, line 29: 👍

mathewjordan (approved):
Let's move it in and then see where things stand.

components/About/CollectionGrid.tsx
adamjarling: Sure.  You mean like grab the Items via an API v2 query?   Query by `id`s for selected items?   Good idea.   I was messing around with some of the `Pres 2` `size` and `region` params, but think I moved past that after making the images background images on elements instead.   I can create a ticket for it.

PR #116, opened 2022-11-16
Apply aspect ratio to Figure.

mathewjordan:
image

## What does this do?

This work brings in an aspect ratio parameter to the Figure component and applies that ratio wherever a figure might be rendered. Currently figures are render on search results, the filter preview modal, and on works for collection cards. 

While applying this:

- Removed LoadingBox package as it didn't seem very helpful for our method paging results.
- Use aspect ratio to build a "placeholder" box while the thumbnail is being requested
- Adjusted transitions Stitches tokens to be digital collections specific
- Adjusted font family Stitches tokens to be Northwestern specific
- Adjusted Grid breakpoints to be derived from media widths


components/Shared/LoadingBox.tsx
```diff
@@ -1,20 +1,26 @@
 import ContentLoader from "react-content-loader";
```
adamjarling, line 1: @mathewjordan Verify we still need this?

adamjarling (approved):

PR #117, opened 2022-11-17
Add Lock Icon to Figure component for restricted works

adamjarling:
## What does this do?
Partial PR to get updates to `<Figure />` component in place.

- Adds styles for the lock icon
- Makes `<Figure />` component aware of a "restricted" state (passed in as `prop`)
- Updates Grid cols to be responsive (so the lock icon doesn't get out of control)

image


types/api/response.ts
```diff
@@ -1,6 +1,6 @@
+import { VisibilityStatus, WorkShape } from "@/types/components/works";
```
mathewjordan, line 1: Oh, forgot this was added. Nice to have here.

mathewjordan (approved):
🚀

PR #118, opened 2022-11-17
Add support for View All Public works Switch on Search page

adamjarling:
## What does this do?
Adds easy access UI support for viewing all public works on the Search page via a Switch/Toggle component.

image

image

## How to test
Toggle the switch on/off from the Search page and notice the `visibility` facet is applied.

### DIFF
--- a/components/Shared/Switch.styled.ts
+++ b/components/Shared/Switch.styled.ts
@@ -1,0 +1,59 @@
+import * as Switch from "@radix-ui/react-switch";
+import { styled } from "stitches.config";
+
+/* eslint sort-keys: 0 */
+
+const StyledSwitch = styled(Switch.Root, {
+  all: "unset",
+  height: "2rem",
+  width: "3.236rem",
+  backgroundColor: "$black10",
+  borderRadius: "9999px",
+  position: "relative",
+  WebkitTapHighlightColor: "transparent",
+  cursor: "pointer",
+
+  "&:focus": {
+    boxShadow: `0 0 0 2px $secondaryAlt`,
+  },
+
+  '&[data-state="checked"]': {
+    backgroundColor: "$darkBlueA",
+    boxShadow: `inset 2px 2px 5px #0003`,
+  },
+});
+
+const StyledThumb = styled(Switch.Thumb, {
+  display: "block",
+  height: "calc(2rem - 12px)",
+  width: "calc(2rem - 12px)",
+  backgroundColor: "$white",
+  borderRadius: "100%",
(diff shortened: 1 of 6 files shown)


components/Shared/Switch.styled.ts
```diff
@@ -0,0 +1,62 @@
...
+  position: "relative",
+  WebkitTapHighlightColor: "transparent",
+  cursor: "pointer",
+
+  "&:focus": {
+    boxShadow: `0 0 0 2px $secondaryAlt`,
+  },
+
+  '&[data-state="checked"]': {
+    backgroundColor: "$purple60",
```
mathewjordan, line 21:
```suggestion
    backgroundColor: "$darkBlueA",
```

components/Shared/Switch.styled.ts
```diff
@@ -0,0 +1,62 @@
+import * as Switch from "@radix-ui/react-switch";
+import { styled } from "stitches.config";
+
+/* eslint sort-keys: 0 */
+
+const StyledSwitch = styled(Switch.Root, {
+  all: "unset",
+  height: "2rem",
+  width: "3.236rem",
+  backgroundColor: "$black20",
```
mathewjordan, line 10:
```suggestion
  backgroundColor: "$black10",
```

components/Shared/Switch.styled.ts
```diff
@@ -0,0 +1,62 @@
...
+  fontSize: "0.8333rem",
+  fontFamily: "$sans",
+  fontWeight: "400",
+  lineHeight: "1em",
+  userSelect: "none",
+  cursor: "pointer",
+  color: "$black50",
+  opacity: "0.7",
+  paddingRight: "0.618rem",
+});
```
mathewjordan, line 58:
```suggestion
const Label = styled("label", {
  fontSize: "$gr3",
  lineHeight: "1em",
  userSelect: "none",
  cursor: "pointer",
  color: "$black50",
  paddingRight: "$gr2",
});
```

mathewjordan (changes requested):
Just a few design ideas... wondering if this is a good chance to bring in a color other than purple?

components/Shared/Switch.styled.ts
mathewjordan: The same blue used for the facet indicator:
image

components/Shared/Switch.styled.ts
mathewjordan: Uses our local stitches tokens for sizing and remove a few things.

mathewjordan (approved):
👍

PR #120, opened 2022-11-22
Case insensitive facet filtering

adamjarling:
## What does this do?
In the Filter modal window, facet filtering is now case insensitive thanks for a helper function provided by @mbklein @kdid and @bmquinn 

image

## How to test
1. From the screen shot above, go to a Work, and open the Filter modal
2. Type facet filter values, and notice case insensitivity in results

### DIFF
--- a/components/Facets/Facet/Options.tsx
+++ b/components/Facets/Facet/Options.tsx
@@ -33,7 +33,13 @@ const FacetOptions: React.FC<FacetOptionsProps> = ({
   });
 
   if (loading) return <SpinLoader />;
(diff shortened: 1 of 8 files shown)


kdid (approved):

adamjarling:
@kdid Good catch on checking quoted filter text input (ie `"Architech"`) not being handled.  Agreed not even sure if this is a valid use case, however now it's implemented and added some tests to the functions building up aggs in queries.  Thanks!

kdid (approved):
👍 

Looks good to me!

PR #121, opened 2022-11-23
Work sliders clean up

adamjarling:
## What does this do?
- Re-orders Work sliders per issue #3342
- Clean up implementation of prepping Work slider collection urls
- Minor styling fix to Collection right hand col image on Work page
- Wrap `BloomIIIF` in an exportable function to avoid multiple dynamic importing across the app.
- Tried a few variations on updating the UI for Collection works/slider/new IIIF Bloom component on the Work page, but no good results.  Will come back to it next week.
- Refactors the Work Type counts for a Collection helper function, and now being used on Work page, Collection List page, and individual Collection page.

Note there's some code in here which doesn't do anything yet, but will be helpful for re-attempting a Work's Collection UI display in the future, I think.


components/Shared/LinkStyled.ts
```diff
@@ -13,6 +13,8 @@ const LinkStyled = styled("a", {
   "&:hover": {
     borderBottom: "1px solid $purple",
   },
-});
+};
 
-export { LinkStyled };
+const LinkStyled = styled("a", linkStyling);
```
mathewjordan, line 18: Oh, wow. Never thought about going at it this way

mathewjordan (approved):
👍 Looks good, maybe we can pair on design/ui next steps?

adamjarling:
> 👍 Looks good, maybe we can pair on design/ui next steps?

Definitely.   That'd help.

PR #124, opened 2022-12-05
Keep SVG to bounds of Icon component.

mathewjordan:
This fixes a bug exhibiting in Safari where paths within SVGs for icons were rendering outside the bounds of their wrapping container. Note the path escaping the wrapping span.

image

Previously:
image
image

Fixed:
image
image


adamjarling (approved):
Looks good.

PR #125, opened 2022-12-06
Style Nectar metadata on Work page

adamjarling:
## What does this do?
- Updates `@samvera/nectar-iiif` package
- Now supports passing a custom delimiter into the Nectar `<Metadata />` component.  We pass in an empty string (to avoid the comma-separated default delimiter), and target the element from client side to get a nice, vertical display of metadata values.

### After
image

### Before
image


mathewjordan (approved):
👍

PR #126, opened 2022-12-09
Search within Collection

adamjarling:
## What does this do?
This supports an option for the user to search either within a Collection, or within All Digital Collections.  Inspired by Github's "Search or jump to..." (look above).

Adds some extra tests around the `<Search />` component.

Even though it's currently just used for expanding a search on the Collection page, it's set up for a `prop` to be passed into the `<Search />` component which could leave room for future implementations outside of the Collection page itself.

image
 
## How to test
- Go to a Collection, and test searching within Collection, or within all the whole site
- Try clicking in/out of the search box to see visibility behavior.
- Go to Home page or main search and notice it doesn't appear.

### DIFF
--- a/components/Header/Primary.test.tsx
+++ b/components/Header/Primary.test.tsx
@@ -1,7 +1,14 @@
 import { render, screen } from "@/test-utils";
 import HeaderPrimary from "./Primary";
 
+import mockRouter from "next-router-mock";
+jest.mock("next/router", () => require("next-router-mock"));
+
 describe("HeaderPrimary", () => {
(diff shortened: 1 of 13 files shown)


davidschober:
This looks really good! It think maybe we can lose the "search this collection button"

mathewjordan (approved):
I somewhat like keeping the search this collection button though the button could using a language change perhaps to `View the Collection` or something less searchy...? The idea is that the button would give the user an entry point into the work items without demanding a search be submitted.

Love it though.

PR #129, opened 2022-12-16
Homepage refinements.

mathewjordan:
image

## What does this do?

This adds two new sections to the homepage for_Works_ and _Collections_ by reusing   `<Grid />` and adding a `<GridFeature />`. Both of these components use the `SearchShape[]` type to render out results.

In the process of doing this, we also now:
- have a cleaned up `page/index.tsx` with a `Hero` pulled out of the layout and Header. 
- have a new component for for `components/Shared/SectionHeading.tsx` 
- have new components for rendering homepage sections in `components/Homepage`


adamjarling (approved):
Looking sharp 👍

PR #131, opened 2022-12-16
Upgrade deps

adamjarling:
General dependency upgrades, including a Bloom upgrade which addresses: https://github.com/nulib/repodev_planning_and_docs/issues/3327


mathewjordan (approved):
Good locally, but looks like a conflict on merge.

PR #134, opened 2022-12-19
Handling display of a Work when in Reading Room

adamjarling:
## What does this do?
Handles display of a Work when in the Reading Room.

#### Authenticated view
image

#### Reading Room view
image

### How to Test
- Run the local dev environment
- In `pages/_app.tsx`, on line 49, set the value of `isReadingRoom` to `true` to mock what the Reading Room auth request response would look like.
```jsx
...
setUser({
  email,
  isLoggedIn,
  isReadingRoom: true,
  name,
  sub,
});
```

### DIFF
--- a/components/Work/ViewerWrapper.styled.ts
+++ b/components/Work/ViewerWrapper.styled.ts
@@ -7,4 +7,16 @@ const ViewerWrapperStyled = styled("section", {
   zIndex: "0",
 });
 
-export { ViewerWrapperStyled };
+const AnnouncementContent = styled("div", {
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+
+  "& svg": {
+    height: "$gr4",
+    fill: "$black80",
+    marginRight: "$gr1",
+  },
+});
+
+export { AnnouncementContent, ViewerWrapperStyled };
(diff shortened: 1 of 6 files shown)


components/Work/ViewerWrapper.styled.ts
```diff
@@ -7,4 +7,16 @@ const ViewerWrapperStyled = styled("section", {
...
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+
+  "& svg": {
+    height: "$gr4",
+    fill: "$black80",
+    marginRight: "$gr1",
+  },
+});
```
mathewjordan, line 20: This looks good. 👨‍🍳

components/Work/ViewerWrapper.test.tsx
```diff
@@ -1,13 +1,50 @@
...
+  user: {
+    email: "joan.doe@northwestern.edu",
+    isLoggedIn: true,
+    isReadingRoom: false,
+    name: "Joan Doe",
+    sub: "jdoe2399",
+  },
+};
+const readingRoomMessage =
+  /You have access to Work because you are in the reading room/i;
```
mathewjordan, line 15: Is this regex?

mathewjordan (approved):
Good from my eyes. That mocked user is something I want to wrap my mind around bit more.

components/Work/ViewerWrapper.test.tsx
adamjarling: Yea, case insensitive to tests don't break if capitalization changes or something.  Not sure this text is 100% locked down.  @davidschober ?

PR #135, opened 2022-12-19
Page transitions slightly smoother and less flashy white

adamjarling:
## What does this do?
Uses the Framer Motion package to attempt handling smooth page transitions.  I can't seem to configure it to do a perfect fade from one page to the next...still get a little bit of the white background feeling, but this is less jarring (I think).

If we don't like this, we can remove the transition altogether and just run with default browser behavior.

## To Test
Click around the site from page to page and bask in its glory.

### DIFF
--- a/components/Transition.tsx
+++ b/components/Transition.tsx
@@ -1,21 +1,23 @@
-import { AnimatePresence, motion } from "framer-motion";
+import { AnimatePresence, type Variants, motion } from "framer-motion";
 import { useRouter } from "next/router";
 
-const variants = {
-  in: {
+const variants: Variants = {
+  animate: {
     opacity: 1,
     transition: {
-      delay: 0.25,
-      duration: 0.25,
+      duration: 0.5,
+      ease: "easeInOut",
     },
-    y: 0,
   },
-  out: {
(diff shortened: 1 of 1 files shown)


mathewjordan (changes requested):
Something about the transitions here feels strange. I can't tell if its  a combination of or any of the `exitBeforeEnter`, `opacity`, `duration`, or the `ease` but it feels like there is a visual kickback, sort of like me driving a manual and not understanding how a clutch works.

mathewjordan:
Maybe we can chat it through? @adamjarling

adamjarling:
Ha.  Yea @mathewjordan no problem.  What I'd like to have happen is the pages are a smooth transition, like one over the other with no opacity.  The part I can't suss out is that the transition effect needs a setting to effect on, like `opacity` for example.   On the beginning, entrance and exit states, if `opacity` is `1` for all, then visually no effect takes place.   Maybe if we had a darker page background it wouldn't be so obvious?   But yea the white just jumps out if we're going this route and not really sure how to get around it.

adamjarling:
Resolved in other work

PR #137, opened 2022-12-20
3405 collection search bugfixes

adamjarling:
## What does this do?
- Fixes some buggy behavior in the Search Within Collections feature (layout, not closing properly, etc)
- Introduces `keydown` mouse event handling (Arrow up, Arrow down, Enter, Escape)
- Breaks the `SearchJumpTo` component out of `Search` to be its own independent component

image

## How to test
- Search within a Collection, experiment and try to break it.


kdid (approved):
👍

PR #138, opened 2022-12-21
Update link to Collection works from Collection page

adamjarling:
## What does this do?
Fixes the Collection page link to view Collection Works.

## How to test
Clicking this link should take you to staging (or production) Search page with the Collection facet activated.

image

### DIFF
--- a/lib/iiif/collection-helpers.ts
+++ b/lib/iiif/collection-helpers.ts
@@ -89,7 +89,7 @@ export const getHeroCollection = (collection: CollectionShape) => {
     return seeAlso;
   };
 
-  const searchUrl = `${DC_URL}/search?q=${id}`;
+  const searchUrl = `${DC_URL}/search?collection=${title}`;
 
   return {
     "@context": "http://iiif.io/api/presentation/3/context.json",
--- a/package-lock.json
+++ b/package-lock.json
@@ -3335,9 +3335,9 @@
       "dev": true
     },
     "node_modules/@types/prettier": {
-      "version": "2.7.1",
-      "resolved": "https://registry.npmjs.org/@types/prettier/-/prettier-2.7.1.tgz",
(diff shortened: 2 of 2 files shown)


kdid (changes requested):
Shouldn't this be `collection.title.keyword`?

lib/iiif/collection-helpers.ts
```diff
@@ -89,7 +89,7 @@ export const getHeroCollection = (collection: CollectionShape) => {
     return seeAlso;
   };
 
-  const searchUrl = `${DC_URL}/search?q=${id}`;
+  const searchUrl = `${DC_URL}/search?collection=${title}`;
```
adamjarling, line 92: @kdid Do you mean this line should be `collection.title.keyword`?   If so, no this line feeds the value `title` which is pulled off the `collection` object here.   

Then this mapper object: https://github.com/nulib/dc-nextjs/blob/76977493c3833c56ceecf2b8fb7358b1e2bdd982/lib/constants/facets-model.ts#L10 connects `collection` to the indexed value `collection.title.keyword`.

Let me know if you want to chat.  You should be able to test it, by following the link to staging.

kdid: Ah, right. I think I meant the first part though. Does `?collection=`  in the query string mean `collection.title.keyword`?  It probably does...?

adamjarling: @kdid Yea, in the facet mapper object below, `id` is what front end code uses as a reference, and `field` is the indexed value we reference in queries.

```
{
    field: "collection.title.keyword",
    id: "collection",
    label: "Collection",
}
```

Which is kinda/sort why I was thrown off by `collection.title.keyword` 2 weeks ago, as it's the only metadata item where "keyword" is included in the faceted value.   Not that it really matters, as long as there's some type of reference able documentation in the future where I lazily don't have to dig through code to see what maps where.

kdid: Right. Looks good. Approving. No docs yet. _ Feel free to start some 😆

kdid (approved):

PR #142, opened 2023-01-03
Block form submit on Collections List filter

adamjarling:
## What does this do?
- Blocks form from submitting (and page refresh) when hitting "enter" key on Collection List page
- Adds a minimum height to `<main />` component (to handle the UI from getting really short when, say no Collection items are returned in this page)
- Some dependency updates

image


mathewjordan (approved):
🚀

PR #143, opened 2023-01-03
Retain metadata filters when re-submitting search values

adamjarling:
## What does this do?
Changes user search behavior slightly, so that metadata filters are retained when updating a search query via `/search?q=something`.  Previously the filters were wiped away on a new search.

## How to test
1. Go to the Search page
2. Apply some Filters
3. Enter a search term
4. Verify that the filters are retained

### DIFF
--- a/components/Search/Search.tsx
+++ b/components/Search/Search.tsx
@@ -8,6 +8,7 @@ import React, {
   useRef,
   useState,
 } from "react";
+import useQueryParams from "@/hooks/useQueryParams";
 import { useRouter } from "next/router";
 
 interface SearchProps {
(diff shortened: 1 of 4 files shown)


components/Search/Search.tsx
```diff
@@ -21,15 +22,14 @@ const Search: React.FC<SearchProps> = ({ isSearchActive }) => {
...
 
   const handleSubmit = (e: SyntheticEvent<HTMLFormElement>) => {
     e.preventDefault();
 
     router.push({
       pathname: "/search",
-      query: {
-        q: searchValue,
-      },
+      query: { q: searchValue, ...urlFacets },
```
mathewjordan, line 32: simple enough!

mathewjordan (approved):

PR #145, opened 2023-01-04
Clean up collection card on a Work.

mathewjordan:
image

## What does this do?

This cleans up the Collection card on a work.

- Renders a real description from `work.collection.description`.
- Adds a read more component to limit long descriptions.
- Tidy up card styling.

## How should we review?

Visit any work. Should be pretty consistent -- though some work/collections might have very short descriptions.


adamjarling (approved):
👍

PR #146, opened 2023-01-05
Consolidate swiper/css imports.

mathewjordan:
image

## What does this do?
This fixes the orientation issue of Bloom sliders by importing the `swiper/css` files in a common location of **layout.tsx**. The orientation issue was caused be a reworking of the Hero component to not be embedded in the Header. When this changed, the `swiper/css` files were no longer available on every page as they had been. Note that the orientation issue needs to be reviewed in a built/deployed Next environment and isn't an issue in dev.

See:
https://preview-3440-bloom-orientation.d1g4r4p7fos4jz.amplifyapp.com/items/8e2d6bc5-3f00-45d1-880b-a458cddf0c1c

## Thoughts?
 I did attempt to also import the necessary CSS files in the `components/BloomWrapper.tsx` file, however jest testing on RelatedItems and SharedLinks components were not fairing well with these CSS files being imported here. Thoughts on this?

### DIFF
--- a/components/Hero/Hero.tsx
+++ b/components/Hero/Hero.tsx
@@ -1,7 +1,3 @@
-import "swiper/css";
-import "swiper/css/effect-fade";
-import "swiper/css/navigation";
-import "swiper/css/pagination";
 import { Autoplay, EffectFade, Keyboard, Navigation, Pagination } from "swiper";
 import { HeroActions, HeroStyled } from "@/components/Hero/Hero.styled";
 import { Label, Summary, Thumbnail } from "@samvera/nectar-iiif";
--- a/components/layout.tsx
+++ b/components/layout.tsx
@@ -1,3 +1,8 @@
+import "swiper/css";
+import "swiper/css/effect-fade";
+import "swiper/css/lazy";
+import "swiper/css/navigation";
+import "swiper/css/pagination";
 import Footer from "@/components/Footer/Footer";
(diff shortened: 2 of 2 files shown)


adamjarling (approved):
@mathewjordan Yea, if this fixes the layout problem, _and_ we're using `swiper`directly in DC v2 (as opposed to just patching an external import, `bloom`), then this makes sense to make the CSS available at a higher level in the tree. 

If we didn't use `swiper` in DC v2, or say another consuming app doesn't use `swiper` directly, will Bloom sliders lay out ok?   So like this like a collision thing?

mathewjordan:
> @mathewjordan Yea, if this fixes the layout problem, _and_ we're using `swiper`directly in DC v2 (as opposed to just patching an external import, `bloom`), then this makes sense to make the CSS available at a higher level in the tree.
> 
> If we didn't use `swiper` in DC v2, or say another consuming app doesn't use `swiper` directly, will Bloom sliders lay out ok? So like this like a collision thing?

Bloom is too depedent on Swiper styling in my opinion. Because of how swiper css styling is imported, Bloom requires you to sideload those css files. So to answer your question, without those CSS files, Bloom would not render well. I think this leans towards moving away from Swiper at some point in the near future. For this reason, I created this issue: https://github.com/nulib/repodev_planning_and_docs/issues/3441

PR #147, opened 2023-01-06
Address Bloom zIndexing and control colors.

mathewjordan:
image

## What does this do?

This work address a lingering zIndex issue on Bloom components by wrapping Bloom with a component having a set `zIndex` of `0`. I also addressed the colors on Bloom controls.

### DIFF
--- a/styles/colors.ts
+++ b/styles/colors.ts
@@ -37,9 +37,19 @@ const basic = {
   gray6: "#f0f0f0",
 };
 
+/**
+ * influence bloom-iiif color tokens
+ */
+const bloom = {
+  accent: `${purple.purple} !important`,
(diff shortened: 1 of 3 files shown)


styles/colors.ts
```diff
@@ -37,9 +37,19 @@ const basic = {
...
 
+/**
+ * influence bloom-iiif color tokens
+ */
+const bloom = {
+  accent: `${purple.purple} !important`,
+  accentAlt: `${purple.purple60} !important`,
+  secondary: `${basic.white} !important`,
+};
+
```
adamjarling, line 48: Nice!

adamjarling (approved):
Nice tight, quick fix.

PR #148, opened 2023-01-06
Hide last-modified manifest metadata item.

mathewjordan:
image

## What does this do?

This simply hides the Last Modified metadata item that is carried in our IIIF Manifests by using the CSS selector rendered in the Nectar component.


components/Work/Metadata.styled.ts
```diff
@@ -10,4 +11,10 @@ const LinkItemStyled = styled("div", {
   },
 });
 
-export { LinkItemStyled };
+const MetadataStyled = styled(Metadata, {
+  "[data-label=last-modified]": {
+    display: "none",
```
adamjarling, line 16: Easy enough for all consumers to use, eh?  Good solution.

adamjarling (approved):

PR #149, opened 2023-01-06
3437 thumb download and share auth bug fix

adamjarling:
## What does this do?
Wraps Download and Share around auth rules.

A logged in user will see this for an "Institution" Work:

image

A non-logged user will see this for an "Institution" Work:

image


components/Work/ActionsDialog/DownloadAndShare.tsx
```diff
@@ -119,7 +150,21 @@ const Item: React.FC<Record<"item", Canvas>> = ({ item }) => {
...
+  if (isValidStringArray) {
+    embedHTMLStringArray[7] = `${width},`;
+    embedHTMLStringArray[9] = `${color}.jpg`;
+  }
+
+  const embedHTMLString = `<img src="${embedHTMLStringArray?.join(
+    "/"
+  )}" alt="Northwestern Libraries Digital Collections Fileset image for ${
+    isValidStringArray ? embedHTMLStringArray[5] : ""
+  }" />`;
```
mathewjordan, line 167: This IIIF image URI construction feels dangerous though I'm not sure of a better way.

mathewjordan (approved):
👍

components/Work/ActionsDialog/DownloadAndShare.tsx
adamjarling: @mathewjordan Ha, agreed.  Brittle and if the shape of the URI ever changes, this needs adjusting.  I couldn't reason about a better way either which didn't involve somewhat complex/confusing JS string parsing.

PR #151, opened 2023-01-09
Update copy text and login for Authenticated Work

adamjarling:
This now looks and acts like:

image


mathewjordan (approved):
Looking good.

PR #153, opened 2023-01-09
Add Academic N and prevent flicker.

mathewjordan:
image

## What does this do?

This adds the Academic N to the fixed header on scroll. I also addressed a flickering issue of the search bar and cleaned up transition in general which may or may not intersect w/ https://github.com/nulib/dc-nextjs/pull/135

### DIFF
--- a/components/Transition.tsx
+++ b/components/Transition.tsx
@@ -1,21 +1,19 @@
 import { AnimatePresence, motion } from "framer-motion";
+import { seconds } from "@/styles/transitions";
 import { useRouter } from "next/router";
 
 const variants = {
   in: {
     opacity: 1,
     transition: {
-      delay: 0.25,
-      duration: 0.25,
+      duration: seconds,
     },
-    y: 0,
   },
   out: {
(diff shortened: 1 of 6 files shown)


components/Transition.tsx
```diff
@@ -1,21 +1,19 @@
 import { AnimatePresence, motion } from "framer-motion";
+import { seconds } from "@/styles/transitions";
```
adamjarling, line 2: Nice

pages/_app.tsx
```diff
@@ -89,7 +92,7 @@ function MyApp({ Component, pageProps }: MyAppProps) {
         })(window,document,'script','dataLayer','GTM-NDJXLQW');
       `}
             </Script>
-            <Component {...pageProps} />
+            {mounted && <Component {...pageProps} />}
```
adamjarling, line 95: 👍

adamjarling (approved):
I like it.  Looks great and implementation feels flexible.  👍

PR #156, opened 2023-01-09
Update Search to bypass dynamic route params

adamjarling:
## What does this do?
Now when searching from a Work page (which has `id` in the `router`'s `query` object from Next/Router), the `Search` component searches for and eliminates `id`, or any other value not in our Facets.   

## How to test
1. Go to a Work page
2. Search for something and notice `id` is no longer in the browser address bar

### DIFF
--- a/components/Search/Search.tsx
+++ b/components/Search/Search.tsx
@@ -8,6 +8,7 @@ import React, {
   useRef,
   useState,
 } from "react";
+import { ALL_FACETS } from "@/lib/constants/facets-model";
 import useQueryParams from "@/hooks/useQueryParams";
 import { useRouter } from "next/router";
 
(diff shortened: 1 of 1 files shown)


components/Search/Search.tsx
```diff
@@ -27,6 +27,9 @@ const Search: React.FC<SearchProps> = ({ isSearchActive }) => {
   const handleSubmit = (e: SyntheticEvent<HTMLFormElement>) => {
     e.preventDefault();
 
+    /* Guard against searching from a Work page, or any dynamic route page */
+    delete urlFacets.id;
```
mathewjordan, line 31: Simple enough. 👍

mathewjordan (approved):

components/Search/Search.tsx
adamjarling: @mathewjordan Changed this up a little bit to be more robust in its checking.

PR #164, opened 2023-01-13
Support downloading of a Work Fileset via download helper package

adamjarling:
## What does this do?
- Prompts a download instead of linking to the IIIF hosted page in a browser
- Updates `@samvera/image-downloader` at https://www.npmjs.com/package/@samvera/image-downloader

image

Will now prompt a download:

image

## How to test
Mimc the above screenshots


mathewjordan (approved):
Works like a charm.

PR #165, opened 2023-01-17
Include Language as a facet

adamjarling:
## What does this do?
-  Supports Language as a facet
- Updates tests to make sure all facets are being rendered to the Filter modal

image


mathewjordan (approved):
🚀

PR #166, opened 2023-01-17
Create WorkCount component and implement.

mathewjordan:
## What does this do?

**Individual Work**
image

**Browse Collections**
image

This creates a `WorkCount` component that can be implemented in any case where we know _audio_, _image_, and _video_ work type counts. To implement in cases this information is passed down to a `Card` figure as a **string**, I allowed the the supplementingInfo prop on the `Card` component to be a `ReactNode` instead, giving greater flexibility at higher levels in our component structure.

## How should we review?

The work impacts three different views:

1. The Collection listing page where each collection item listing renders `WorkCount`.
2. Any work page where we deliver a `Card` rendering `WorkCount` for the related collection.
3. Any work page where the actions dialog for **Find**, **Cite**, and **Download & Share** are rendered. In this case the `WorkCount` component is not rendered and instead a react fragment wraps what a string for `WorkType` of the specific image.

### DIFF
--- a/components/Shared/WorkCount/WorkCount.tsx
+++ b/components/Shared/WorkCount/WorkCount.tsx
@@ -1,0 +1,67 @@
+import { IconAudio, IconImage, IconVideo } from "@/components/Shared/SVG/Icons";
+import {
+  WorkCountTotal as Total,
+  WorkCountType as Type,
+  WorkCountTypes as Types,
+  WorkCountStyled,
+} from "@/components/Shared/WorkCount/WorkCount.styled";
+import React from "react";
+
+export interface WorkCountProps {
+  audio?: number;
+  image?: number;
+  video?: number;
+}
+
+const WorkCount: React.FC<WorkCountProps> = ({
+  audio = 0,
+  image = 0,
+  video = 0,
+}) => {
+  const total = audio + image + video;
+
+  return (
+    <WorkCountStyled>
+      <Total data-testid="work-count-total">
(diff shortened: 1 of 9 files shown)


components/Shared/WorkCount/WorkCount.tsx
```diff
@@ -0,0 +1,67 @@
...
+}) => {
+  const total = audio + image + video;
+
+  return (
+    <WorkCountStyled>
+      <Total data-testid="work-count-total">
+        {total} {total !== 1 ? "Works" : "Work"}
+      </Total>
+      <Types>
+        {image ? (
```
adamjarling, line 29: Clever on the coercions!  👍

components/Work/ActionsDialog/Aside.tsx
```diff
@@ -12,13 +12,15 @@ const ActionsDialogAside: React.FC<ActionsDialogAsideProps> = ({
...
+  const WorkType = () => <>{work?.work_type}</>;
+
   return (
     <ActionsDialogAsideStyled data-testid="actions-dialog-aside">
       {work && (
         <Card
           title={work.title}
           imageUrl={work.thumbnail}
-          supplementalInfo={work.work_type}
+          supplementalInfo={<WorkType />}
```
adamjarling, line 23: Nice

adamjarling (approved):
Really good work, both visually and code clarity.  Love it.

PR #167, opened 2023-01-18
New Search results pagination

adamjarling:
## What does this do?
- Cleans up pagination to simplify it, and be more maintainable.
- Saved the old one just in case
- Add tests to make sure proper pagination results are returned

image

image

## How to test
Experiment with using pagination on Search page

### DIFF
--- a/components/Search/Pagination.styled.ts
+++ b/components/Search/Pagination.styled.ts
@@ -26,7 +26,7 @@ const prevNextStyles = {
 };
 
 /* eslint sort-keys: 0 */
-export const PaginationStyled = styled("nav", {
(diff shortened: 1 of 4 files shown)


mathewjordan (approved):
Wow, I really like this. Doing away with numbers cleans things up a lot. 😻 

My only question is wheether we would want to still render but disable the Next and Previous buttons on _first_ and _last_ pages respectively. The button moving on location caught me for a second. Could be an iterative thing after discussing with folks? I don't think that should stop us from going forward and pushing this through though.

adamjarling:
@mathewjordan Yea, good idea.  I'll disable the buttons instead of hiding them.

PR #168, opened 2023-01-19
Address filter button styling, update transitions.

mathewjordan:
## What does this do?
This handles a few different issues.

- Addresses Filter button padding issue on the dropdown.
- Updates some transitions that never made the changeover from the `$all` to `$dcAll` token
- Fixes a small bug in the `/collections` list page where the Jump To search component was accidentally rendering there.
- Adjust styling of the Jump To after a quick conversation with @adamjarling 

## How should we review?
- Glance at search, scroll up and down, add some facets, remove some facets
- Glance at `/collections`, attempt a search
- Glance at a Collection page, attempt a search


adamjarling (approved):
Looking good

PR #170, opened 2023-01-19
Add Rights Statement facet link to Work metadata

adamjarling:
## What does this do?
Adds the new Rights Statement metadata as a faceted link

image


mathewjordan (approved):
So simple. 🥇

PR #171, opened 2023-01-19
Adjust Nectar metadata formatting to include none search metadata.

mathewjordan:
## What does this do?

image

This reworks some previously completed code used for customizing Links to faceting from metadata to wrap all values in a stylized bulleted `div`. If the `label` also has  `searchParam`, a `next/link` will wrap the metadata value. Otherwise a `span` will.

### Notes
- I use the `dangerouslySetInnerHTML` prop to render values as some values do contain HTML (Related URL)
- I'm not entirely in love with this approach as it does require curated constants for formatting. Long-term, we may want to look into better extending out some options relating to `customValueContent` in the Nectar metadata component so that this is not required.

### DIFF
--- a/components/Work/Metadata.test.tsx
+++ b/components/Work/Metadata.test.tsx
@@ -1,6 +1,5 @@
-import WorkMetadata, { ValueAsSearchLink } from "@/components/Work/Metadata";
+import WorkMetadata, { ValueAsListItem } from "@/components/Work/Metadata";
 import { render, screen, within } from "@/test-utils";
-import { FACETS_WORK_LINK } from "@/lib/constants/works";
 import { MetadataItem } from "@iiif/presentation-3";
 import { manifest } from "@/mocks/sample-work-image";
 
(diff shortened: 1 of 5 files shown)


components/Work/Metadata.test.tsx
```diff
@@ -24,11 +23,15 @@ describe("WorkMetadata component", () => {
     });
   });
 
-  FACETS_WORK_LINK;
-
   it("renders metadata value as custom link pattern", () => {
-    render(<ValueAsSearchLink param="foo" value="bar" />);
+    render(<ValueAsListItem searchParam="foo" value="bar" />);
```
adamjarling, line 27: I like `searchParam` better, was curious about that before...

lib/constants/works.ts
```diff
@@ -1,46 +1,94 @@
...
+  },
+  {
+    label: "Table of Contents",
   },
   {
     label: "Technique",
-    param: "technique",
+    searchParam: "technique",
   },
 ];
```
adamjarling, line 92: This reads much cleaner to me of the intent.  Dig it.

adamjarling (approved):
This adds a lot of clarity.  Can always refactor later if you want, but I like it and makes sense. 👍

PR #172, opened 2023-01-19
Add Sitemap to the app

adamjarling:
## What does this do?
Manually adds a link to our generated sitemap file.

## How to test
- In local dev mode: verify this url exists as a crawl-able, XML file: [URL]/sitemap.xml
- On staging, verify this url exists as a crawl-able, XML file: https://dc-next.rdc-staging.library.northwestern.edu/sitemap.xml

### DIFF
--- a/public/sitemap.xml
+++ b/public/sitemap.xml
@@ -1,0 +1,9 @@
+<!-- public/sitemap.xml -->
+   <xml version="1.0" encoding="UTF-8">
+   <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+     <url>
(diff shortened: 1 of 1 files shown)


mathewjordan (approved):
Nice and easy.

kdid (approved):

davidschober:
Will this work if we have more than one sitemap.xml.gz file? Google says we have to break them into chunks at 50k? The initial one then references the other sitemaps recursively.

adamjarling:
@davidschober I'm not sure...guessing the pointer logic is inside the sitemap file itself?   We should definitely test and see.

PR #173, opened 2023-01-19
Wire Honeybadger into the app

adamjarling:
## What does this do?
Adds Honeybadger code to the Next app.  @mbklein added the supporting Terraform vars to staging and production.

## How to test?
1. In a local dev environment, throw a `new Error("ima error")` in some component. 
2. Log into https://app.honeybadger.io/projects and view the `dc-next` project and the error should appear.

### DIFF
--- a/lib/honeybadger/config.js
+++ b/lib/honeybadger/config.js
@@ -1,0 +1,47 @@
+import Honeybadger from "@honeybadger-io/js";
+
+const setupHoneyBadger = () => {
+  // https://docs.honeybadger.io/lib/javascript/reference/configuration.html
+  const sharedHoneybadgerConfig = {
+    apiKey: process.env.HONEYBADGER_API_KEY,
+    environment: process.env.HONEYBADGER_ENV || process.env.NODE_ENV,
+    projectRoot: "webpack://_N_E/./",
+
+    // Uncomment to report errors in development:
+    reportData: true,
+
+    revision: process.env.AWS_COMMIT_ID,
+  };
+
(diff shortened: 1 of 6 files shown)


lib/honeybadger/config.js
```diff
@@ -0,0 +1,46 @@
+import Honeybadger from "@honeybadger-io/js";
+
+const setupHoneyBadger = () => {
+  // https://docs.honeybadger.io/lib/javascript/reference/configuration.html
+  const sharedHoneybadgerConfig = {
+    apiKey: process.env.HONEYBADGER_API_KEY,
+    environment: process.env.NODE_ENV,
```
mbklein, line 7: Recommending we change this to
```suggestion
    environment: process.env.HONEYBADGER_ENV || process.env.NODE_ENV,
```
so that we can override it if we want without monkeying with the `NODE_ENV` variable.

lib/honeybadger/config.js
```diff
@@ -0,0 +1,46 @@
...
+    projectRoot: "webpack://_N_E/./",
+    revision: process.env.HONEYBADGER_REVISION,
```
mbklein, line 9: The way to get the revision differs in Amplify than it did when building under GitHub Actions. I believe it should be
```suggestion
    revision: process.env.AWS_COMMIT_ID,
```

mbklein (changes requested):
The suggested environment variable name changes from `lib/honeybadger/config.js` will also need to be reflected in `next.config.js`.

mbklein (approved):

PR #175, opened 2023-01-20
Format number util for large numbers to read as 12,345 etc...

adamjarling:
## What does this do?
Introduces a helper function to format `12345` like:

```js
formatNumber(12345);  // 12,345
```

and a helper function to display nouns and suffixes properly.

```js
pluralize("bear", 3);  // 3 bears
```


adamjarling:
I'm going to combine this PR with #3476

mathewjordan (approved):
Awesome to have these. Was thinking they could totally be in the same helper file as well.

PR #176, opened 2023-01-20
Address Safari CSS anomalies.

mathewjordan:
### Before
image

### After
image

This does quite a bit more that just address SVG issues related to safari. While auditing, I noticed a couple of other things that I decided to address as well.

- Addresses SVGs on Filter Modal to now show properly.
- Addresses Safari transition bug on the Filter button when scrolling where the Filter button did not center properly
- Addresses Safari font rendering where Safari would be VERY bold if Northwestern brand fonts were used in conjunection with `font-weight: 700` 
- Addresses a strange fixed positioning bug where the Dropdown of UserFacets on the Filter button if it is currently opened and the use scrolled to the top by forcing the dropdown to close
- General styling updates!

## How should we review?

-  Tryout the search page in Safari and your preferred browser
- Toggle filter, add a few facets, scroll around, in general just play with things
- Things should like a bit neater in general should be much smoother

### DIFF
--- a/components/Facets/Facet/GenericFacet.styled.ts
+++ b/components/Facets/Facet/GenericFacet.styled.ts
@@ -10,6 +10,7 @@ const StyledGenericFacet = styled("div", {
   [`${StyledHeading}`]: {
     margin: "0 0 1rem",
     fontFamily: "$northwesternDisplayBold",
+    fontSize: "$gr5",
     fontWeight: "400",
   },
 });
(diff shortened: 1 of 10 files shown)


adamjarling (approved):
@mathewjordan Looks like quite a bit of updates.  Works smooth as butter in Safari, feels like nice, tight responsive UX.  Great improvements.

PR #177, opened 2023-01-24
Homepage works query from keywords match on "featured" 

mathewjordan:
image

## What does this do?

This simply now queries to the API endpoint via a POST method, and requests a random set of works matching the `keywords` field with the term `featured`. All randomizing and limiting **(25)** occurs in the OpenSearch query. No constants for this are need and this query should provide us with some set of works for the appropriate environment.

## How should we review?

Pretty straight forward at this but but...

1) How does the query look?
2) Any unnecessary noise?


components/Homepage/Works.tsx
```diff
@@ -1,10 +1,80 @@
...
+                  keywords: "featured",
+                },
+              },
+              random_score: {},
+            },
+          },
+          size: size,
+        };
+
+        const response = await axios.post(DC_API_SEARCH_URL, body);
```
adamjarling, line 46: @mathewjordan This is going to need `withCredentials: true` as an `axios` config, for our queries to respect auth user cookies.

Which gets me thinking about the issue I'm working on... (converting everything to `axios`), and the `lib/dc-api.ts` helper.  Instead of putting `axios` requests directly in components, we could wrap them all in little services (ie. `getCollection(id)`, `getHomePageWorks()`, `getApiData(...)` etc)?   And then if we do expand @kdid 's API Client, our wrapper functions can easily be identified and extracted?

lib/constants/homepage.ts
```diff
@@ -349,569 +348,3 @@ export const featuredCollections = [
...
-      "https://dcapi.rdc-staging.library.northwestern.edu/api/v2/works/e7cf42de-ee57-49ea-bf95-aa855a3e7129/thumbnail",
-    visibility: "Public",
-    iiif_manifest:
-      "https://dcapi.rdc-staging.library.northwestern.edu/api/v2/works/e7cf42de-ee57-49ea-bf95-aa855a3e7129?as=iiif",
-    work_type: "Image",
-    id: "e7cf42de-ee57-49ea-bf95-aa855a3e7129",
-    title:
-      "Sam Hinton and Alice Stuart, Folk Song Jamboree at Stern Grove, 1964",
-  },
-];
```
adamjarling, line 917: Always warms my heart to see ~600 lines of code trimmed 😁

adamjarling (commented):
@mathewjordan So remind me how `featured` hooks in?  A user sets this on a Work in Meadow?

adamjarling (approved):

mathewjordan:
> @mathewjordan So remind me how `featured` hooks in? A user sets this on a Work in Meadow?

Yeah, it's a `string` value added to the Keywords field in meadow.

mathewjordan:
> Looking around, we're using `axios` directly in other components (Search for example). Approving this, and will create another ticket to wrap all API Search Post requests under a service module.

YESSS! We should absolutely do this.

PR #179, opened 2023-01-24
3484 homepage updates

adamjarling:
## What does this do?
- Updates the Homepage featured Collections to use the curated list received.
- Refactors Homepage featured Collections to pull in dynamically instead of hard coded
- Introduces `HomepageContext` to store data which doesn't change during an app session
- Refactors the `GridItem` slightly to use the API `thumbnail` endpoint as image source, or... an enhanced, hi-res version of the image if being used by a consumer component which needs bigger sizing, better resolution, etc.
- Wraps `GridItem` with a unit test

image

### DIFF
--- a/components/Grid/Item.tsx
+++ b/components/Grid/Item.tsx
@@ -1,5 +1,4 @@
 import { useContext, useEffect, useState } from "react";
-import { DCAPI_ENDPOINT } from "@/lib/constants/endpoints";
 import Figure from "@/components/Figure/Figure";
 import { GridItem as ItemStyled } from "@/components/Grid/Grid.styled";
 import Link from "next/link";
@@ -8,11 +7,11 @@ import { UserContext } from "@/pages/_app";
 
 interface GridItemProps {
   item: SearchShape;
+  isFeatured?: boolean;
 }
 
-const GridItem: React.FC<GridItemProps> = ({ item }) => {
+const GridItem: React.FC<GridItemProps> = ({ item, isFeatured }) => {
   const [urlPath, setUrlPath] = useState<string>();
-  const [endpointPath, setEndpointPath] = useState<string>();
   const [supplementalInfo, setSupplementalInfo] = useState<string>();
   const userContext = useContext(UserContext);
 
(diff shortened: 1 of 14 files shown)


components/Grid/Item.tsx
```diff
@@ -26,32 +25,29 @@ const GridItem: React.FC<GridItemProps> = ({ item }) => {
...
       <Link href={`${urlPath}/${item.id}`}>
-        <a>
+        <a data-testid="grid-item-link">
           <Figure
             data={{
               aspectRatio: item.representative_file_set.aspect_ratio,
               isRestricted: isRestricted(item),
-              src: `${DCAPI_ENDPOINT}${endpointPath}/${item.id}/thumbnail`,
+              src: isFeatured
+                ? `${item.representative_file_set.url}/square/512,/0/default.jpg`
```
mathewjordan, line 49: Would this work? It'd route every Item/Figure src URI through our API endpoint, maybe clear things up for future usage with `next/image`? Not critical, just a thought.

```suggestion
                ? `${item.thumbnail}?aspect=square`
```

components/Homepage/Overview.tsx
```diff
@@ -18,8 +18,10 @@ const HomepageOverview = () => {
           <Content>
             <h2>Enrich your research with primary sources</h2>
             <p>
-              Explore millions of high-quality primary sources and images from
-              around the world, including artworks, maps, photographs, and more.{" "}
+              Explore digital resources from the Northwestern University Library
```
mathewjordan, line 21: So glad to see the JSTOR language gone. 🥳

lib/homepage-helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+    const workIds = collections.map((c) => c.representative_image.work_id);
+    const worksResponse = await axios.all(
+      workIds.map((id) => axios.get(`${DCAPI_PRODUCTION_ENDPOINT}/works/${id}`))
+    );
+    const works: WorkShape[] = worksResponse.map(({ data: { data } }) => ({
+      ...data,
+      representative_file_set: {
+        ...data.representative_file_set,
+        /** Format for nice UI display on the home page */
+        aspect_ratio: 1,
```
mathewjordan, line 28: 👍

mathewjordan (approved):
I think this is solid. See my one suggestion, but take it or leave it. 🚀

lib/homepage-helpers.ts
adamjarling: Nice call @mathewjordan on the easy handle

PR #182, opened 2023-01-26
Numerical string sorting of Series agg values.

mathewjordan:
## What does this do?

This creates a numerical string sorting helper that can be applied to an array of Agg results following the shape of
 
```ts
{
  key: string;
  doc_count: number;
}
```

This helper is then applied to the `series` array that renders the accordion UI on the Collection Organization page for each individual collection.

The expect behavior is that this will sort arrays by the **key** `string`, and apply a compare  function to ensure numeric strings are ordered as expected, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare

And thanks to @bmquinn !!!

## How should we review?
Spin up DC Next. Take a look at the ABC.123.TEST collection in staging. The Collection Organization series should be ordered as expected. A test for sort function also has been written.

### DIFF
--- a/lib/collection-helpers.ts
+++ b/lib/collection-helpers.ts
@@ -321,3 +321,14 @@ export async function getTopMetadataAggs({
     return [];
   }
 }
+
+/**
+ * Specialized array helper for numerical string
+ * sorting an array by the `key` property
+ */
+export const sortAggsByKey = (arr: GenericAggsReturn[]) => {
+  const collator = new Intl.Collator("en", { numeric: true });
+  return arr.sort(function (a, b) {
+    return collator.compare(a.key, b.key);
+  });
+};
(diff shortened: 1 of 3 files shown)


lib/collection-helpers.ts
```diff
@@ -321,3 +321,14 @@ export async function getTopMetadataAggs({
     return [];
   }
 }
+
+/**
+ * Specialized array helper for numerical string
+ * sorting an array by the `key` property
+ */
+export const sortAggsByKey = (arr: GenericAggsReturn[]) => {
+  const collator = new Intl.Collator("en", { numeric: true });
```
adamjarling, line 330: This is cool.  I also had no clue this ever existed.  Pretty handy.  👍

adamjarling (approved):

PR #185, opened 2023-01-27
Cleanup getStaticPaths for dynamic pages.

mathewjordan:
## What does this do?

This little bit of work just cleans out some old code that is no longer being used to generate static paths.

The current version of DC Next does not generate paths on build for `items/[id]` or `collections/[id]`, and instead relies on the `fallback` to create static pages through SSR when those routes are requested. For this reason, getStaticProps on these routes is still somewhat beneficial.

Also, the older function calls of `getWorkIds` and `getCollectionIds`  were not actually returning responses from our current DC API. These requests were built around an early mocked API response in API Gateway -- and I think should be considered trash at this point.

## How should we review?

Seem reasonable? Long-term, I'm wondering how much of this can be moved to some server side stuff in a Next 13 `app` directory?

### DIFF
--- a/lib/collection-helpers.ts
+++ b/lib/collection-helpers.ts
@@ -17,29 +17,6 @@ export async function getCollection(
   }
 }
 
-export async function getCollectionIds(): Promise<Array<string>> {
-  const body = {
-    _source: ["id"],
-    query: {
-      query_string: {
-        query: "*",
-      },
-    },
-    size: 1000,
-  };
-
-  try {
-    const response = await apiPostRequest<ApiSearchResponse>({
-      body,
-      url: `${process.env.NEXT_PUBLIC_DCAPI_ENDPOINT}/search/collections`,
-    });
-    return response?.data ? response?.data.map((item) => item.id) : [];
-  } catch (err) {
-    console.error("Error getting all Collection Ids", err);
-    return [];
-  }
-}
-
 export async function getCollectionWorkCount(collectionId: string) {
   const body = {
     _source: ["id"],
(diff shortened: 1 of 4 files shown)


lib/collection-helpers.ts
```diff
@@ -40,50 +40,6 @@ export async function getCollectionList(
...
-    if (response?.aggregations) {
-      return response.aggregations.allIds.buckets.map((bucket) => bucket.key);
-    }
-    return [];
-  } catch (err) {
-    console.error("Error getting all Collection Ids", err);
-    return [];
-  }
-}
-
```
adamjarling, line 86: @mathewjordan I refactored this function as it wasn't doing it's job.  We're using it to build the Collection pages dynamically... but that's probably going away in the next issue I pick up converting this request to Client side...

lib/work-helpers.ts
```diff
@@ -17,48 +17,6 @@ export async function getWork(id: string) {
...
-      method: "GET",
-      url: `${DC_API_SEARCH_URL}/works/${id}?as=iiif`,
-    });
-    return response;
-  } catch (err) {
-    console.error("Error getting the work", id);
-    return null;
-  }
-}
-
```
adamjarling, line 61: Same thing I think for now, until we switch over from `getStatic....` we still need this no?

pages/collections/[id].tsx
```diff
@@ -139,12 +138,9 @@ const Collection: NextPage<CollectionProps> = ({
...
 export async function getStaticPaths() {
-  const ids = await getCollectionIds();
-  const paths = ids.map((id) => ({ params: { id } }));
-
   return {
     fallback: "blocking",
-    paths,
+    paths: [],
   };
 }
```
Comment (adamjarling) [inline pages/collections/[id].tsx:146]: Ah, now I think I get it.  Does this work then passing an empty array but still grabbing data with `getStaticProps`?  Like, it will work with ISR?

adamjarling (approved):
Looks great.

PR #187, opened 2023-01-27
Clean up network requests

adamjarling:
## What does this do?
1. Replaces all instances of `fetch` with `axios`
2. Creates some new helpers for `GET` and `POST` requests
3. Replaces TypeScript type definitions for some manual request typing with official Types from the `elasticsearch` library
4. Creates a User Context, and remove from `_app.tsx`
5. Lots of little clean up along the way

## How to test?
Run through the site in a browser, and keep an eye out for any edge case bugs.


lib/collection-helpers.ts
```diff
@@ -1,83 +1,39 @@
...
-    size: 0,
+    size: 1000,
   };
 
   try {
-    const response = await getAPIData<ApiSearchResponse>({
+    const response = await apiPostRequest<ApiSearchResponse>({
       body,
-      url: `${process.env.NEXT_PUBLIC_DCAPI_ENDPOINT}/search`,
+      url: `${process.env.NEXT_PUBLIC_DCAPI_ENDPOINT}/search/collections`,
```
mathewjordan, line 34: I removed this function `getCollectionIds()` in #185  as I don't think we need it anymore NOR did it do anything, but trying to understand what this route does...

lib/dc-api.ts
```diff
@@ -1,42 +1,75 @@
...
+}
+
+interface ApiPostRequestParams {
+  body?: ApiSearchRequestBody;
   url: string;
 }
 
-async function getAPIData<R>(obj: GetApiParams): Promise<R | undefined> {
-  const { body, method = "POST", url } = obj;
+async function apiGetRequest<R>(
```
mathewjordan, line 13: 💯 these two functions are going to really help clean things up.

package.json
```diff
@@ -47,6 +47,7 @@
     "swiper": "^8.3.2"
   },
   "devDependencies": {
+    "@elastic/elasticsearch": "7.17",
```
mathewjordan, line 50: 👍

pages/_app.tsx
```diff
@@ -1,21 +1,15 @@
-import { User, UserContextInterface } from "@/types/context/user";
 import type { AppProps } from "next/app";
```
mathewjordan, line 1: Nice cleanup of _app.

mathewjordan (approved):
Everything seems great and the right directional move to make for a healthier codebase.

lib/collection-helpers.ts
adamjarling: @mathewjordan I think at one point it existed to get Collection `id`s?   That route allows you to search in Collections, and not Works (which is default at the `/search` route...I think.  @kdid could probably verify.

kdid: Yes. That's it.

PR #188, opened 2023-01-30
Include auth in API requests coming from DC v2 client

adamjarling:
## What does this do?

Move network requests client side for Work, Collection and Collection List.  Assuming Reading Room will now work that auth is being passed through.

## How to test
Verify from Reading Room

### DIFF
--- a/components/Collection/Item/Item.tsx
+++ b/components/Collection/Item/Item.tsx
@@ -4,7 +4,7 @@ import {
   ItemImageWrapper,
   ItemStyled,
 } from "@/components/Collection/Item/Item.styled";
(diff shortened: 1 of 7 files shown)


kdid (approved):
The API change for the `/collections/id` route went through and I'm seeing private collections now on your branch.

Approving. I am still seeing that Openseadragon appendChild error frequently on this branch but other than that looks good to me.

adamjarling:
@kdid Regarding #3522 above,  after rebasing I'm not seeing the "appendChild" error nearly as much.  Like 1 in 20 attempts... but yea let's get this code introduced for the bigger Auth win, and can address the loading errors if they keep popping up.

PR #192, opened 2023-01-31
First pass at using dcapi-types

kdid:


adamjarling:
@kdid There's some merge conflicts here, and after briefly reviewing I might just start a clean branch.  One quick comment for structure, I think we tuck `dcapi-types` under the `@nulib` NPM Organization.  So the import will look like:

`import { ... } from '@nulib/dcapi-types';`

PR #193, opened 2023-02-01
Resolve query infinite growth issues for search results and aggs.

mathewjordan:
## What does this do?

This fixes a pretty ugly issue relating to queries on search results and aggs infinitely growing and not being able to be destructed. The entire `buildQuery()` call has now been recreated to ensure that the a new query is built from scratch on each unique request.

```tsx
export function buildQuery(obj: BuildQueryProps) {
  const { aggs, aggsFilterValue, size, term, urlFacets } = obj;

  const must: QueryDslQueryContainer[] = [];
  if (term) must.push(buildSearchPart(term));
  if (Object.keys(urlFacets).length > 0) must.push(addFacetsToQuery(urlFacets));

  return {
    ...querySearchTemplate,
    ...(must.length > 0 && {
      query: {
        bool: {
          must: must,
        },
      },
    }),
    ...(aggs && { aggs: buildAggs(aggs, aggsFilterValue, urlFacets) }),
    ...(typeof size !== undefined && { size: size }),
  }
```

### DIFF
--- a/lib/queries/builder.ts
+++ b/lib/queries/builder.ts
@@ -1,6 +1,7 @@
 import { buildSearchPart, querySearchTemplate } from "@/lib/queries/search";
 import { ApiSearchRequestBody } from "@/types/api/request";
 import { FacetsInstance } from "@/types/components/facets";
+import { QueryDslQueryContainer } from "@elastic/elasticsearch/api/types";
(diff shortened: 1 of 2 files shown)


adamjarling (approved):
🙏 Way, way better.

PR #195, opened 2023-02-01
Update width on fileset download.

mathewjordan:
## What does this do?

Oh, this is simple. Really simple. The IIIF image URI now just asks for a size of `3000,` instead of the previous 1000.


adamjarling (approved):
Simple enough. 🤷

PR #196, opened 2023-02-01
Add alt from canvas label if it exists.

mathewjordan:
## What does this do?

This is a pretty simple fix the adds an `alt` to the Necter rendered canvas thumbnail from the `label`. If a label does not exist, we fallback to the `id`. This takes advantage of the `altAsLabel` prop in the Nectar `Thumbnail`.


adamjarling (approved):
👍

PR #198, opened 2023-02-01
Hide duplicate series title, extend bloom styling.

mathewjordan:
## What does this do?

image

This extends the Related Items stitches styling to the wrap the section around the Collection Organization accordions, and then hides the Bloom render homepage and label elements for those Bloom instances.


adamjarling (approved):
👍

PR #200, opened 2023-02-01
Wrap social icons with working buttons.

mathewjordan:
## What does this do?

This wraps our existing social icons with working buttons using the `react-share` package that is also used on the legacy digital-collections.


adamjarling (approved):
Looks like a good solution for now.  Nice.

PR #204, opened 2023-02-02
Update to use @nulib/dcapi-types

adamjarling:
## What does this do?
In general, transitions from using manually defined TypeScript types in DC v2, to using types generated from `OpenSearch` and exposed via the `@nulib/dcapi-types` package.   This syncs the source of truth for what valid Indexed data is.

## How to test
Pretty much just walk through the entire app and notice anything which breaks.  (This sync should pretty significantly reduce the surface area of where bugs can appear).


lib/collection-helpers.ts
```diff
@@ -1,25 +1,22 @@
...
-import { VisibilityStatus } from "@/types/components/works";
+import { type Visibility } from "@nulib/dcapi-types";
 import { shuffle } from "@/lib/utils/array-helpers";
 
 export type CollectionListShape = {
-  description?: string;
+  description?: string | null;
   id: string;
-  representativeImage: CollectionRepresentativeImage;
+  representativeImage: Collection["representative_image"];
```
kdid, line 12: There is a `CollectionRepresentativeImage` exported, would you want to use it here?
https://github.com/nulib/dcapi-types/blob/main/index.d.ts#L4

kdid (approved):

mathewjordan (approved):
Love it. Definitely the move in the right direction for our codebase.

PR #205, opened 2023-02-02
Update Clover to support HLS requests withCredentials.

mathewjordan:


mbklein (approved):
Working great from my dev environment

PR #207, opened 2023-02-03
Swap out Facet aggs request useFetchApiData hook for apiPostRequest.

mathewjordan:
## What does this do?

This swaps out the older `geFetchApiData` in favor the more current `apiPostRequest()`.

Along with this we set the default value for **searchTerm** that is peeled off of `router.query.q` to be undefined. There was a React timing bug that sometimes allowed through the default value of an empty string even if a term was set. We now do not make aggregation requests unless we have verified the value of `router.query.q`. This seems to have fixed the root issue here.

## How should we review?

- Go to `/search`
- Complete a search for `Berkeley`
- Hit filter, check agg doc_counts,
- They should remain consistent as the open and close the filter modal.

### DIFF
--- a/components/Facets/Facet/Options.tsx
+++ b/components/Facets/Facet/Options.tsx
@@ -1,8 +1,16 @@
+import {
+  ApiResponseAggregation,
+  ApiSearchResponse,
+} from "@/types/api/response";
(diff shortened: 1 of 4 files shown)


kdid (approved):
Looks 👍 to me

PR #209, opened 2023-02-03
Bloom options updates, disable preview.

mathewjordan:
## What does this do?

This work updates the Bloom IIIF package, allowing us to disable the preview option on the Bloom slider. Additionally, I cleaned up the options been sent to each Bloom instance: set credentials to `include`,  customized the breakpoints to match our constant breakpoints, and set an explicit false on the **enablePreview** property.

## How should we review?
Go to any work. Bloom sliders should render without previews on hover/focus for each item. Breakpoints should also match our breakpoints better.

### DIFF
--- a/components/BloomWrapper.tsx
+++ b/components/BloomWrapper.tsx
@@ -1,5 +1,9 @@
+import { SwiperProps } from "swiper/react";
 import dynamic from "next/dynamic";
+import { gr } from "@/styles/sizes";
+import { rem } from "@/styles/global";
 import { styled } from "@/stitches.config";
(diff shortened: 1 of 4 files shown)


components/BloomWrapper.tsx
```diff
@@ -12,15 +16,62 @@ const StyledBloomIIIFWrapper = styled("div", {
   },
 });
 
+type SwiperBreakpoints = SwiperProps["breakpoints"];
```
adamjarling, line 19: Nice

adamjarling (approved):

PR #214, opened 2023-02-06
Update mobile styling for Work Filter

adamjarling:
## What does this do?
Tightens up mobile display for Work Filter modal:

image

image


mathewjordan (approved):
That'll do for now -- and makes it a lot more usable. Would love to come back to this at some point and think how we can make it my "appy" and investigate a few good examples. 🚀

PR #215, opened 2023-02-06
Create responsive menu.

mathewjordan:
## What does this do?

This adds a global minimal responsive menu at the `@sm` breakpoint. State is handled locally in the `<Super/>` subcomponent of `<Header>`

image


adamjarling (approved):
Looks good to me.

PR #216, opened 2023-02-06
Update Collection page for mobile / tablet responsive

adamjarling:
## What does this do
Cleans up the Collection page (but not Search part of page...yet).

image


mathewjordan (approved):
Nice and tight. 🎸

PR #217, opened 2023-02-06
3313 work modals

adamjarling:
## What does this do?
Updates Work modals (Find / Cite / Download) to behave better on mobile/tablet.

### Before
image

### After
image


components/BloomWrapper.tsx
```diff
@@ -14,6 +14,18 @@ const StyledBloomIIIFWrapper = styled("div", {
   ".swiper-slide[data-type='collection']": {
     display: "none",
   },
+
+  "@sm": {
+    "& > div > div": {
```
mathewjordan, line 19: Good for now but man I hate this div cascade. Some near future tech debt to come in and make these targets more specific. Also could just be a Bloom thing too?

mathewjordan (approved):

PR #220, opened 2023-02-07
Clean up homepage for responsive

adamjarling:
## What does this do?
Makes the homepage look much better for responsive.


mathewjordan (approved):
✅

PR #232, opened 2023-02-09
Add no results message.

mathewjordan:
## What does this do?

This adds a nice visually centered no results message on the `/search`page. Pagination also does not show any longer if results are `0`. Most of the work here is styling with some minor reworking of the conditionals to render the Results messages and the Pagination. 

image


adamjarling (approved):
Looks great.

PR #233, opened 2023-02-09
Implement next/image for Figures, add LQIP.

mathewjordan:
## What does this do?

This work leverages `next/image` for use in all Figure components. This will affect Figures on search results pages, the homepage, the filter modal, and the Collection card on work pages.

While implementing, I also added a LQIP (Low Quality Image Placeholder) with dimensions of `3px` using the `size` parameter on the thumbnail endpoint. These LQIP images will load in prior to the higher quality thumbnails. Once the the thumbnail has loaded, the opacity will transition to create a Medium like effect. I opted for this method rather leaning on https://nextjs.org/docs/api-reference/next/image#blurdataurl as this requires base64 encoding and would complicate things a bit more than necessary.

## How should we review?

Take a look at search results, move back and forth through paging.

### DIFF
--- a/components/Figure/Figure.styled.ts
+++ b/components/Figure/Figure.styled.ts
@@ -1,19 +1,21 @@
 import * as AspectRatio from "@radix-ui/react-aspect-ratio";
 import { VariantProps, styled } from "@/stitches.config";
 import { IconLock } from "@/components/Shared/SVG/Icons";
+import Image from "next/image";
 
 /* eslint sort-keys: 0 */
 
 IconLock.toString = () => ".icon-lock";
 
-const FigureImage = styled("img", {
+const FigureImage = styled(Image, {
   display: "flex",
   borderRadius: "3px",
-  transition: "$dcAll",
+  transition: "$dcImageLoad",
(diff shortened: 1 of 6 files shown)


components/Figure/Figure.styled.ts
```diff
@@ -1,19 +1,21 @@
...
 import { VariantProps, styled } from "@/stitches.config";
 import { IconLock } from "@/components/Shared/SVG/Icons";
+import Image from "next/image";
 
 /* eslint sort-keys: 0 */
 
 IconLock.toString = () => ".icon-lock";
 
-const FigureImage = styled("img", {
+const FigureImage = styled(Image, {
```
adamjarling, line 10: Cool that this works for `Image`

adamjarling (changes requested):
@mathewjordan Possible to remove these warnings?

image

All in all it's super slick and feels great in the UX!

mathewjordan:
@adamjarling Good catch. I've added `sizes` that should quiet things down.

adamjarling (approved):

PR #234, opened 2023-02-10
Add an API route and rewrites to fetch sitemaps from an S3 bucket

mbklein:
This PR contains all the changes needed to deliver sitemaps from the root of the server without having to trigger a build every time the sitemap changes.

- Create a public-read static website bucket to hold sitemaps
- Create a NextJS api route `/api/sitemap/:filename` to stream content from the sitemap bucket to the client
- Add three rewrite rules to the Amplify app to transparently rewrite `/sitemap*` to `/api/sitemap/sitemap*` without redirecting the client

Results can be seen in the preview branch at https://preview-sitemap-bucket.d2v1qbdeix3nr2.amplifyapp.com/sitemap.xml

There is a [companion PR](nulib/meadow#3225) to tell Meadow about the new sitemap location and have it generate the proper index file

### DIFF
--- a/lib/constants/bucket.ts
+++ b/lib/constants/bucket.ts
@@ -1,0 +1,3 @@
+const DC_SITEMAP_BUCKET = process.env.NEXT_PUBLIC_DC_SITEMAP_BUCKET;
+
+export { DC_SITEMAP_BUCKET };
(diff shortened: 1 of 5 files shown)


adamjarling (approved):
Looks good.  This feels more "right" to me than trying to shoehorn it into a Next page route.  Nice work. 👍

PR #237, opened 2023-02-13
GTM dataLayer send user logged in status

adamjarling:
## What does this do?
Fixes GTM's `dataLayer` user logged in status.

image

## How to test
From any page in DC, load the page, then in the browser console type `dataLayer` to get access viewing the `window` object.  Look for `isLoggedIn`, being logged in, and not being logged in.


mathewjordan (approved):
Awesome, love the new helper.

PR #240, opened 2023-02-13
Responsive refinements.

mathewjordan:
## What does this do?

This work resolves a handful of small responsive issues that have lingered for a bit.

  - The Clear All and Public Works Only actions now collapse into columns in a small viewport. We also have show the Work Type actions as well (these were hidden). Transitions on scrolling to a fixed search bar are now quite a bit smoother here too. 
  - The zIndex issue on the Search bar for Image works is resolve. The fixed search bar now glides over top of the viewer, rather than under.
  - Some small font size adjustments were made to increase legibility in small viewports on an individual work
  - The search bar has some minor styling adjustments, 1) the placeholder text now has an ellipsis for overflow, 2) the clear button for search queries renders with a fading gradient over text inputs, 3) the search button is purple and pops a bit more, and 4) some minor margin/padding updates were made.
  - Bloom label and summary are now centered on `xs` viewports.

### DIFF
--- a/components/BloomWrapper.tsx
+++ b/components/BloomWrapper.tsx
@@ -15,9 +15,10 @@ const StyledBloomIIIFWrapper = styled("div", {
     display: "none",
   },
 
-  "@sm": {
+  "@xs": {
     "& > div > div": {
       alignItems: "center",
+      textAlign: "center",
 
       "& > div": {
         "&:last-child": {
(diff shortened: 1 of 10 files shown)


adamjarling (approved):
These changes look really, really good.  Nice to see mostly styling updates.  Great work!

PR #242, opened 2023-02-15
Fix anchor button nesting issue on homepage causing console warnings

adamjarling:
## What does this do
- Fixes warning on homepage in the browser console (a `button` was nested under an `a` tag)

## How to test
Open the homepage locally and view browser console.  Notice no warnings.


mathewjordan (approved):
👍

PR #243, opened 2023-02-16
Remove fuzziness and wildcard from Search

adamjarling:
## What does this do?
Updates Search to eliminate fuzziness and get more accurate results.   It allows for direct pass through searches for "power searching" as well.

This is part one of being able to add future auto suggest and "did you mean..." capabilities.

## How to test
Try the following searches and they should return expected results (previously returned no results).

```
title:(carol AND kathy)
_exists_:title
NOT _exists_:title
NOT title:(carol OR kathy)
```


mathewjordan (approved):
Seems to simplify things from a power user endpoint. I like it.

PR #244, opened 2023-02-17
Tell `next/image` not to proxy and optimize.

mathewjordan:
## What does this do?

This updates our `next/image` components to not attempt to optimize the src files. Previously Next was funneling our original `src` thumbnails through some optimization process and outputting them as `.webp` files. We should now be going direct to source for these and not having a ambiguous Next middleware handling this. File size on the user end will be a little larger but we should no longer have the reading room issue.

Checkout the preview branch build here:
https://preview-3675-reading-room-thumbnails.d2v1qbdeix3nr2.amplifyapp.com/search

Before: (current Production)
image

After: ([preview branch](https://preview-3675-reading-room-thumbnails.d2v1qbdeix3nr2.amplifyapp.com/search))
image

### DIFF
--- a/components/Figure/Figure.tsx
+++ b/components/Figure/Figure.tsx
@@ -51,6 +51,7 @@ const Figure: React.FC<FigureProps & FigureVariants> = (props) => {
             sizes={srcSetSizes}
             src={lqip.toString()}
             priority={true}
+            unoptimized={true}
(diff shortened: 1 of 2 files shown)


adamjarling:
Untested on my end, but code looks hopeful 😁

PR #247, opened 2023-02-20
Add basic 404.

mathewjordan:
## What does this do?

This work adds a basic custom 404 page. The page is basically just reusing the Hero component on the homepage to render a single slide. Additionally, it adds navigation to the end user to direct them back to actual content.

The image itself can be easily swapped out by adjusting a IIIF thumbnail on the **pageNotFoundCollection** object at `lib/constants/404.ts` if we decided on different one.

I also fixed a small bug in the HeroBasic component where the `bgImage` prop does not drill down as expected to some inline CSS. This was notice when working through an initial idea.

image

### DIFF
--- a/components/Hero/Basic.styled.ts
+++ b/components/Hero/Basic.styled.ts
@@ -10,6 +10,7 @@ const HeroBasicStyled = styled("section", {
   display: "flex",
   justifyContent: "center",
   alignItems: "center",
+  background: "url(/images/liz__O8A9903_final.jpg)",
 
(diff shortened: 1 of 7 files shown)


adamjarling (commented):
Looks and works great.  One thing to maybe consider?  Say I'm at a Work page, or Collection page and mistype the `id` or any dynamic route id.   Should the 404 kick in, or some note that the `id` item doesn't exist?   Currently it returns this...

image

PR #249, opened 2023-03-07
Updates related to Lighthouse feedback

adamjarling:
## What does this do?
Makes some SEO improvements to the page based on feedback from Lighthouse.  Most of the improvement suggestions were related to dependent packages.  Within this code base, the 2 primary updates were...

1. Pulling in `next/font` and using local fonts instead of grabbing from NU CDN.
2. Pulling in the Bundle Analyzer package to help identify JS weight and bundle size.

## To Test
Feel free to run the Bundle Analyzer, Lighthouse in general, and make sure there are no odd font-related display issues.


context/user-context.tsx
```diff
@@ -1,8 +1,11 @@
 import React, { ReactNode, createContext, useState } from "react";
-import { type User, type UserContext } from "@/types/context/user";
+import {
+  type User,
+  type UserContext as UserContextType,
+} from "@/types/context/user";
 import { getUser } from "@/lib/user-helpers";
 
-const UserContext = createContext<UserContext>({ user: null });
+const UserContext = createContext<UserContextType>({ user: null });
```
mathewjordan, line 8: Good catch.

mathewjordan (approved):
Everything looks really good. Things keep getting smoother and smoother here. I really like the font change. 🚀

PR #252, opened 2023-03-09
Remove all references to useFetchApi hook

adamjarling:
## What does this do
Removes a troublesome hook `useFetchApiData()` previous used to get API /search endpoint data.   

## How to test
Go to the search page and interact with facets/filtering, etc.  Everything should behave as expected in the UI.


mathewjordan (approved):
This is stellar. Glad to have this hook out of there.

PR #256, opened 2023-03-15
Upgrade dependencies and fix Type errors which are fixable

adamjarling:
## What does this do?
Upgrades some major dependency versions, and fixes some long-standing type errors.


mathewjordan (approved):
🚀

PR #257, opened 2023-03-15
Only include Reading Room message on restricted Works. Update README

adamjarling:
## What does this do?
Only displays the "You're see this in the Reading Room because....etc" message when a Work is restricted.

This also reverts the server-side 404 check for Work and Collection pages, to allow access in Reading Room.  With the server side check (no auth), the API was returning no items (as it should) for auth protected works and collections.

## How to test
1. Log into a Reading Room machine via Microsoft Remote Desktop
2. View both a Public and Institution (or Private) Work

Notice the message, or absence of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,19 @@
-# Digital Collections v2 NextJS App
+# Digital Collections v2
 
-This is a work in progress, for experimenting with NextJS, AWS Amplify and various ways of building / data fetching / hosting.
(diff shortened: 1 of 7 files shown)


mathewjordan (commented):
👍 for the message on Restricted works in reading room.

🤔  I'm hesitant about breaking status codes for this exception. The route `/items/foo` is a 200. Of course the 404 page is a 404 which is great, but before we roll with this, can we get together to chat through it a bit more? Perhaps with @davidschober too?

image

mathewjordan (approved):
Let roll!

PR #258, opened 2023-03-17
Only include Reading Room message on restricted Works. Update README

mathewjordan:


mathewjordan:
Created for review of #257

PR #262, opened 2023-03-20
Update DC API types.

mathewjordan:
## What does this do?

This brings dc-nextjs up to date for the `@nulib/dcapi-types` package. Some mocks and tests required updating, otherwise nothing big.


adamjarling (approved):
👍

PR #273, opened 2023-04-26
Restricted access email link includes Work Id

adamjarling:
## What does this do?
Replaces the email link (currently just copies the email address), to a hyperlink with `mailto:` which populates the email "Subject" with a message: "Work ID: XYX access request".

image

image

## How to test
Click on the email link for a Private Work (when not logged in), and it should open your email client with Subject pre-filled.


mathewjordan (approved):
👍

PR #275, opened 2023-05-02
Address encoding of subject within the related collection helper.

mathewjordan:
## What does this do?

This patches an issue where the subject was not being properly URI encoded when being appended to the the Collection URI. While, I feel like this entire helper could use some cleanup, I opted for the quick fix in encoding subject.

See: https://devbox.library.northwestern.edu:3000/items/ef5f341f-50cc-4ff3-a84c-7f13ac0d1afa where the Sliders render even though the subject "Adam & Eve" is present.

image

### DIFF
--- a/lib/work-helpers.ts
+++ b/lib/work-helpers.ts
@@ -34,17 +34,17 @@ export function getWorkSliders(work: Work) {
 
   /** Collection slider */
   if (work.collection) {
-    const collectionLabel = work.collection.title;
+    const encodedCollectionLabel = encodeURIComponent(work.collection.title);
     const collectionSummary = `Collection`;
     workSliderUrls.push(
-      `${DC_API_SEARCH_URL}?query=collection.id:"${work.collection.id}"&collectionLabel=${collectionLabel}&collectionSummary=${collectionSummary}&as=iiif`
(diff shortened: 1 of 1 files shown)


lib/work-helpers.ts
```diff
@@ -55,8 +55,9 @@ export function getWorkSliders(work: Work) {
     const subjects = shuffle(work?.subject.map((s) => s.label)).slice(0, 2);
 
     subjects.forEach((subject: string) => {
+      const encodedSubject = encodeURIComponent(subject);
```
adamjarling, line 58: Our good friend, `encodeURIComponent`.  👍

adamjarling (commented):
@mathewjordan Should we apply this fix also to everywhere we're building the `url` in this helper file?  ie.

```jsx
`${DC_API_SEARCH_URL}?query=collection.id:"${work.collection.id}"&collectionLabel=${collectionLabel}&collectionSummary=${collectionSummary}&as=iiif`

// Updated url

`${DC_API_SEARCH_URL}?query=collection.id:"${work.collection.id}"&collectionLabel=${encodeURIComponent(collectionLabel)}&collectionSummary=${encodeURIComponent(collectionSummary)}&as=iiif`
```

If we're passing values via query params, the parsing will always hiccup on unencoded `&` ampersands, no?

adamjarling (approved):
Looking good, and this will get us by no problem.

PR #276, opened 2023-05-03
Add copy link to download and embed assets.

mathewjordan:
## What does this do?

This adds a Copy IIIF button with a IIIF `svg` to each image fileset item in the download and embed modal. When clicked, the info.json URL (https://iiif.stack.rdc-staging.library.northwestern.edu/iiif/2/73d4f5fd-1f8e-4699-85ec-92bc4351ba6a/info.json) will be copied to the user's clipboard.

image


adamjarling (approved):
Looks sharp.  Nice. 🙏

PR #286, opened 2023-06-22
Update message for opening a work in Mirador for authenticated Works

adamjarling:
## What does this do?
Removes the link and adds a warning message to Institution or Private Works in the Download and Share modal for a Work.

image

## How to test
1. Search for an Institution Work and observe the message and inactive link.

### DIFF
--- a/components/Work/ActionsDialog/DownloadAndShare.tsx
+++ b/components/Work/ActionsDialog/DownloadAndShare.tsx
@@ -15,7 +15,6 @@ import {
   ShareURL,
   ShareURLActions,
 } from "@/components/Work/ActionsDialog/DownloadAndShare.styled";
-
 import { Label, Thumbnail } from "@samvera/nectar-iiif";
 import React, { useEffect, useState } from "react";
 import {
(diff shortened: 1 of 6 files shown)


components/Work/ActionsDialog/DownloadAndShare.tsx
```diff
@@ -98,13 +117,22 @@ const DownloadAndShare: React.FC = () => {
...
               </ShareURLActions>
+              {(isWorkInstitution || isWorkPrivate) && (
+                <Announcement
+                  css={{
+                    marginTop: "1rem",
+                  }}
+                  data-testid="mirador-announcement"
+                >
+                  Opening in external tools like Mirador not supported for works
+                  that require authentication
```
mathewjordan, line 133:
```suggestion
                  Opening in external tools like Mirador is not supported for works
                  that require authentication.
```

mathewjordan (changes requested):
Just suggestion on a typo. Looks good otherwise.

mathewjordan (approved):
Looking great. Let's go. 🚀

PR #287, opened 2023-06-23
Add support for License facet

adamjarling:
## What does this do
Supports `license` as a linkable facet field for metadata on a Work.   

License will also appear in the Filter Dialog on the Search page.

image

## How to test
1. Go to Search page and notice License is now available as a Filter option.
2. On a Work page, if `license` is set for a Work, it will appear as a clickable link in the Work's metadata display (Update to DC API v2 must be live on Staging first)

### DIFF
--- a/lib/constants/works.ts
+++ b/lib/constants/works.ts
@@ -43,6 +43,10 @@ const WORK_METADATA_LABELS: WorkMetadata[] = [
     searchParam: "language",
   },
   {
+    label: "License",
+    searchParam: "license",
+  },
+  {
     label: "Location",
   },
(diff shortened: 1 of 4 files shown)


lib/constants/works.ts
```diff
@@ -42,6 +42,10 @@ const WORK_METADATA_LABELS: WorkMetadata[] = [
     label: "Language",
     searchParam: "language",
   },
+  {
+    label: "License",
+    searchParam: "license",
```
mathewjordan, line 47: Cool, forgot how this works.

PR #289, opened 2023-08-09
Update Viewer configuration on /embedded-viewer route to display the meta…

adamjarling:
## What does this do?
Shows the metadata side panel in a Viewer on a `/embedded-viewer` item route.

image

## How to test
1. In DC go to any Work and click "Download and Share", and then copy the `<iframe src` attribute value as seen below:

image

2. Copy this URL into a browser window.
3. Notice the metadata panel in the Viewer is open
4. View a regular Work in DC, and notice the metadata panel in the Viewer is not open.


davidschober:
@adamjarling This was supposed to be for the /embed. Did I mess up the issue?

adamjarling:
@davidschober Nope.  I assumed you meant '/shared` for some reason.  Updated to work for the `/embedded-viewer` route.

kdid (commented):
👍

PR #290, opened 2023-08-11
Test clover suite

adamjarling:
## What does this do?

Pulls in the new Clover Release `@samvera/clover-iiif@latest`

## How to test
Fire up the app and view the Work page.  Note the Viewer, Metadata, and Related Items.  

Also view Collection pages or anywhere else a `Slider` component may be used.


adamjarling:
@mathewjordan This one is now good to go with `latest` Clover release.   Tests passing.

components/Clover/SliderWrapper.tsx
```diff
@@ -1,13 +1,13 @@
+import Slider from "@samvera/clover-iiif/slider";
```
mathewjordan, line 1: Liking the moving of this to a `Clover` directory. Should we also move the **ViewerWrapper** as well at some point?

mathewjordan (approved):
Works like a dream.

components/Clover/SliderWrapper.tsx
adamjarling: @mathewjordan Yes, good idea.  I'll do that before the squash.

PR #292, opened 2023-08-29
Support displaying "similar to a Work" search results

adamjarling:
## What does this do?
When a user clicks on the "View All" button in a Clover Slider on a Work page, now they'll see Works which are "similar", "related to", etc.

image

image

- The `/search` page now keeps track of a new query param, `similar`, and if it exists a UI will render noting the Search page is showing Works "similar" to a particular Work.   
- A user can still interact with all facets as normal, and the Works should filter down accordingly.
- Dismissing the "similar" alert will remove the filter on search results.  This is a one-way action and cannot be restored from the UI on the current page.  A user will have to go to another Work page and click "view all" from another Slider to re-initiate.

## How to test
1. Go to a Work page and click "view all" button on a "More Like This" Slider.
2. Note the new "similar" UI.
3. Explore adding / removing facets.


package.json
```diff
@@ -35,7 +35,7 @@
     "@radix-ui/react-select": "^1.0.0",
     "@radix-ui/react-switch": "^1.0.1",
     "@radix-ui/react-tabs": "^1.0.1",
-    "@samvera/clover-iiif": "^2.0.0",
+    "@samvera/clover-iiif": "^2.0.2",
```
mathewjordan, line 38: :tada

PR #294, opened 2023-09-05
4155 embedded viewer query params

adamjarling:
## What does this do?
Supports user control of displaying certain Clover `Viewer` elements for embedded players.   The user can supply query params at the end of the embedded URL, which will get picked up by the embedded player and communicated to Clover `Viewer`.

Here are the current Clover Viewer config options:

```ts
export type ViewerConfigOptions = {
	background?: string;
	canvasBackgroundColor?: string;
	canvasHeight?: string;
	ignoreCaptionLabels?: string[];
	informationPanel?: {
		open?: boolean;
		renderAbout?: boolean;
		renderSupplementing?: boolean;
		renderToggle?: boolean;
	};
	openSeadragon?: OpenSeadragonOptions;
	requestHeaders?: IncomingHttpHeaders;
	showIIIFBadge?: boolean;
	showTitle?: boolean;
	withCredentials?: boolean;
};
```

This issue supports `showTitle`, `informationPanel.open` and `informationPanel.renderToggle`.

One caveat is that the query params must be `Uri Encoded`.  The way the embedding display is wired-up, we can't just let the user append something like `url....?showTitle=false&informationPanelOpen=true` Unless we trust a user to correctly manually `encodeUri` the params, we need some type of helper feature from which they can copy/paste.  So I built some select drop downs which will update the iFrame `src`, and drafted up a quick visual:

image

image

Also as part of this issue I broke up the `DownloadAndShare` component into more manageable sub-components.

## How to test?
1. Go to a Work page and click "Download and Share"
2. In the Embed Viewer helper, select different options
3. Copy and past the `<iframe src=[COPY ME] />` url and drop that into a fresh browser tab.  The route should respect and render your selected options.


components/Work/ActionsDialog/DownloadAndShare/EmbedViewer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    return params.join("&");
+  }
+
+  function handleShowInfoPanelChange(e: React.ChangeEvent<HTMLSelectElement>) {
+    switch (e.target.value) {
+      case "show open":
+        setInfoPanelDefaultOpen("true");
+        setShowInfoPanel("true");
+        break;
+      case "show closed":
```
mathewjordan, line 73:
```suggestion
      case "show default closed":
```

components/Work/ActionsDialog/DownloadAndShare/EmbedViewer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    if (showInfoPanel) params.push(`informationPanel=${showInfoPanel}`);
+    if (infoPanelDefaultOpen)
+      params.push(`informationPanelOpen=${infoPanelDefaultOpen}`);
+    if (showTitle) params.push(`showTitle=${showTitle}`);
+    return params.join("&");
+  }
+
+  function handleShowInfoPanelChange(e: React.ChangeEvent<HTMLSelectElement>) {
+    switch (e.target.value) {
+      case "show open":
```
mathewjordan, line 69:
```suggestion
      case "show default open":
```

mathewjordan: It looks like this was not matching the values set in the options.

components/Work/ActionsDialog/DownloadAndShare/EmbedViewer.tsx
mathewjordan: Same as above.

pages/embedded-viewer/[manifestId].tsx
```diff
@@ -14,9 +16,49 @@ const EmbeddedViewerPage: NextPage = () => {
...
+        informationPanel === "true" && informationPanelOpen === "true"
+          ? true
+          : false,
+      renderAbout:
+        informationPanel === "true" && informationPanelAbout === "true"
+          ? true
+          : false,
+      renderToggle: informationPanel === "true" ? true : false,
+    },
+    showIIIFBadge: showIIIFBadge === "true" ? true : false,
```
Comment (mathewjordan) [inline pages/embedded-viewer/[manifestId].tsx:58]: What if we just set this to false since this isn't a customizable option. Might be a little less confusing to end users.
```suggestion
    showIIIFBadge: false,
```

pages/embedded-viewer/[manifestId].tsx
```diff
@@ -14,9 +16,49 @@ const EmbeddedViewerPage: NextPage = () => {
...
+      {} as { [key: string]: string }
+    );
+  console.log("filteredSearchParams", filteredSearchParams);
+
+  const {
+    informationPanel = "true",
+    informationPanelOpen = "false",
+    informationPanelAbout = "true",
+    showTitle = "true",
+    showIIIFBadge = "true",
```
Comment (mathewjordan) [inline pages/embedded-viewer/[manifestId].tsx:42]: Or rather, maybe this should just be false as the default here?

mathewjordan (changes requested):
Just a minor bug related to a value mismatch. Otherwise looks and works great.

adamjarling:
> Just a minor bug related to a value mismatch. Otherwise looks and works great.

Yea, would be pretty straight forward to do.  Was kinda testing the waters on whether this visual is even necessary, or clunky?   @davidschober any thoughts?

components/Work/ActionsDialog/DownloadAndShare/DownloadAndShare.styled.ts
```diff
@@ -111,6 +111,12 @@ const ItemStyled = styled("div", {
   margin: "$4 0 0",
 });
 
+const LayoutImg = styled("img", {
+  width: "100%",
+  opacity: "0.8",
+  marginBottom: "$gr3"
+});
+
```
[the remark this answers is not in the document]
adamjarling: Yea, would be pretty straight forward to do. Was kinda testing the waters on whether this visual is even necessary, or clunky? @davidschober any thoughts?

adamjarling:
@mathewjordan Updated with some clean up, simplifying the logic around what the user can actually configure in the UI.  And built in the 'real time' preview to match `select` element options.  And added a test to the main component here.

mathewjordan (approved):
Let's roll with it.

PR #295, opened 2023-09-11
Debug GTM VirtualPageView events, add `collections` data to `/collections/*` routes.

mathewjordan:
## What does this do?

In the end, this doesn't do too much, but does:

1. add the previously omitted title to a `/collections/*` route dataLayer payload.
2. add basic defensiveness around the dataLayer.push, checking for a page title
3. breaks up the dataLayer.push into two steps so that the `event` for only comes after data has been defined

Steps 2 and 3 will be watched over the next few days to test their effect.

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -7,6 +7,7 @@ import {
   camptonExtraBold,
   camptonExtraLight,
 } from "@/styles/fonts";
+
 import type { AppProps } from "next/app";
 import Head from "next/head";
 import { ObjectLiteral } from "@/types";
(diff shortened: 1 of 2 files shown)


pages/_app.tsx
```diff
@@ -44,19 +45,27 @@ function MyApp({ Component, pageProps }: MyAppProps) {
...
-        event: "VirtualPageView",
-        // @ts-ignore
-        ...pageProps.dataLayer,
-        isLoggedIn: user?.isLoggedIn,
-      };
+      const { dataLayer } = pageProps;
+
       // @ts-ignore
-      window.dataLayer?.push(payload);
+      if (dataLayer?.pageTitle) {
```
adamjarling, line 53: Probably makes sense, since we'll likely never have a page (Collection, Work, About, etc) without a `Title` in production?  Or is this a possibility?  @davidschober ?

PR #297, opened 2023-09-13
Update Clover, fixing minor accessibility issues.

mathewjordan:
- This small fix brings the accessibility work from #4070 into Digital Collections.
-  It also incorporates the alignment issue of the Clover Viewer options if a `showTitle` is false and the Toggle or IIIF Badge are present.
- Finally, more clean up on the GTM dataLayer, removing an unused deconstructed const.


adamjarling (approved):
👍

PR #306, opened 2023-10-10
Update Readme

adamjarling:
## What does this do?
Updates the README to note our strategy for keeping NextJS >= 1 version behind `next@latest`. The idea is this will give some buffer space between AWS Amplify catching up to NextJS's latest features.

@mbklein When making note of pinned dependencies, curious if we really need to pin Honeybadger?


mbklein:
`@honeybadger-io/js` had a breaking change that I don't think we ever coded for. It's definitely worth getting current if we can. I'm not even sure we need the base package if we just include `@honeybadger-io/react` as a dependency and use it properly.

mbklein (approved):

PR #310, opened 2024-02-02
Integrate chat component on search page as-is.

mathewjordan:
This work brings previously developed code for the Chat component (formerly SearchPrototpe) from the homepage to search results screen.

See in Amplify:
https://preview-4479-integrate-chat.d2v1qbdeix3nr2.amplifyapp.com/

Note that line 13 below is hardcoded as the `/chat-endpoint` path does not yet exist on our staging API.

https://github.com/nulib/dc-nextjs/blob/3e38afcf3004cc81b5075f995ea055236028c989/components/Chat/components/Wrapper.tsx#L10-L22


.husky/pre-commit
adamjarling, line 1: Are we removing these checks for a reason?  Probably good to keep consistent w/ regular DC pre-commit checks.

adamjarling (approved):
@mathewjordan See the one comment I made about removing the pre-commit hook, but other than that looks good to me.

.husky/pre-commit
mathewjordan: @adamjarling

PR #313, opened 2024-03-04
Build out Search bar components and state

adamjarling:
Refactor of the components making up the Search bar.

image


mathewjordan (approved):
This looks excellent.

PR #318, opened 2024-03-14
Retain search query param between route searches

adamjarling:
## What does this do?
When initiating a search from any route in the application, if the "Use Generative AI" checkbox is checked, the Search route will now retain this value and immediately kick into AI Search.  Also retains value on page refreshes.

For code changes, it makes the following adjustments:
- Removes user selection of "use generative ai" from Search Context, and moves this to a query param in URL
- In general cleans up previous implementations of extracting and applying facet values as query params

This seems like it smooths out the UX.  If we want to take it further, we could push tracking user selection of "use generative ai" to `localStorage`.

## How to test
1. Go to any route except an individual Collection page.  
2. Check "Use Generative AI", enter a search term and submit.
3. Search page should immediately kick off the AI Search

### DIFF
--- a/components/Facets/Filter/Submit.tsx
+++ b/components/Facets/Filter/Submit.tsx
@@ -17,7 +17,7 @@ const FacetsFilterSubmit: React.FC<FacetsFilterSubmitProps> = ({
   const {
     query: { q },
   } = router;
-  const { urlFacets } = useQueryParams();
+  const { ai, urlFacets } = useQueryParams();
 
   const {
     filterDispatch,
@@ -34,6 +34,7 @@ const FacetsFilterSubmit: React.FC<FacetsFilterSubmitProps> = ({
       ...(q && { q }),
       ...urlFacets,
       ...userFacetsUnsubmitted,
+      ...(ai && { ai }),
     };
 
     router.push({
--- a/components/Facets/UserFacets/UserFacets.tsx
+++ b/components/Facets/UserFacets/UserFacets.tsx
@@ -1,4 +1,5 @@
 import * as Dropdown from "@radix-ui/react-dropdown-menu";
+
 import {
   DropdownContent,
   DropdownToggle,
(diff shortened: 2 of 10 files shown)


components/Facets/Filter/Submit.tsx
```diff
@@ -17,7 +17,7 @@ const FacetsFilterSubmit: React.FC<FacetsFilterSubmitProps> = ({
   const {
     query: { q },
   } = router;
-  const { urlFacets } = useQueryParams();
+  const { ai, urlFacets } = useQueryParams();
```
mathewjordan, line 20: ✨

components/Facets/UserFacets/UserFacets.tsx
```diff
@@ -93,7 +97,11 @@ const FacetsCurrentUser: React.FC<FacetsCurrentUserProps> = ({
       if (screen === "search")
         router.push({
           pathname: "/search",
-          query: { ...(q && { q }), ...newObj },
+          query: {
+            ...(q && { q }),
+            ...newObj,
+            ...(ai && { ai }),
```
mathewjordan, line 103: Does order here matter?

components/Search/Search.tsx
```diff
@@ -36,22 +37,27 @@ const Search: React.FC<SearchProps> = ({ isSearchActive }) => {
   const handleSubmit = (e: SyntheticEvent<HTMLFormElement>) => {
     e.preventDefault();
 
+    const updatedFacets: UrlFacets = {};
+
     /* Guard against searching from a page with dynamic route params */
-    const facetIds = ALL_FACETS.facets.map((facet) => facet.id);
+    const allFacetsIds = getAllFacetIds();
```
mathewjordan, line 43: Nice.

types/context/search-context.ts
```diff
@@ -2,7 +2,6 @@ import { ApiResponseAggregation } from "@/types/api/response";
 
 export interface SearchContextStore {
   aggregations?: ApiResponseAggregation;
-  isGenerativeAI: boolean;
```
mathewjordan, line 5: 👏

mathewjordan (approved):
I like it a lot. Keeps `ai` in line with how we handle other search relevant values, but breaks it out as special.

components/Facets/UserFacets/UserFacets.tsx
adamjarling: @mathewjordan I don't think it does, in terms of where `ai=` lives in the query string.  As long as we don't have a Facet named `ai`, we should be fine. 😀

PR #324, opened 2024-05-02
Refactor/experiment with chat socket component

adamjarling:
## What does this do?
After briefly experimenting with [socket.io](https://socket.io/), looks like it needs to be implemented on both the client and server ends.  Wish I would have discovered that earlier 😀.  

The end result of this PR does:

- Re-organizes web socket connection logic/handling back into the `useChatSocket()` hook
- Provides a convenience `message` and `sendMessage` helper which a consuming component (or components) can more easily use.
- Eliminates the Chat Wrapper component we originally put in place to handle timing issues of the web socket's ready state

Visually, everything should look the same in the UI.

## How to test?  
Just a basic run through of expected functionality.  I'd try to break it, by interrupting searches, page refreshes, etc.  The connection will throw errors, but the UI should not crash.


hooks/useChatSocket.ts
```diff
@@ -1,11 +1,17 @@
...
 import { DCAPI_CHAT_ENDPOINT } from "@/lib/constants/endpoints";
+import { StreamingMessage } from "@/types/components/chat";
 import axios from "axios";
 
 const useChatSocket = () => {
-  const [chatSocket, setChatSocket] = useState<WebSocket | null>(null);
   const [authToken, setAuthToken] = useState<string | null>(null);
+  const [url, setUrl] = useState<string | null>(null);
+
+  const socketRef = useRef<WebSocket | null>(null);
```
mathewjordan, line 11: I'm like this convention of using refs instead of state. Makes me think we can do this in many more situations.

PR #326, opened 2024-05-07
Use neural search with AI Chat searching

adamjarling:
## What does this do?
Updates the OpenSearch query object to use a `hybrid` search (using `neural`), when a user is searching via AI Chat.

## How to test?
*Note it's a little tough to test UI search results against limited staging data*

- Update your local `miscellany` repo for env variables needed for `hybrid` search
- Perform a regular and Chat AI search.  
- Inspect the request body, and notice when using AI Chat search (if a search term is provided), the request payload should resemble something like:
- The Search results should visually appear as expected, but the paginated results will be calculated differently by the API

```jsx
{
   ...
    "query": {
        "hybrid": {
            "queries": [
                {
                    "bool": {
                        "must": [
                            {
                                "query_string": {
                                    "fields": [
                                        "title^5",
                                        "all_text",
                                        "all_controlled_labels",
                                        "all_ids"
                                    ],
                                    "query": "Do we have music by Mozart?"
                                }
                            }
                        ]
                    }
                },
                {
                    "neural": {
                        "embedding": {
                            "k": 40,
                            "model_id": NEXT_PUBLIC_OPENSEARCH_MODEL_ID,
                            "query_text": "Do we have music by Mozart?"
                        }
                    }
                }
            ]
        }
    },
    "size": 40
}
```

### DIFF
--- a/components/Facets/Filter/Modal.tsx
+++ b/components/Facets/Filter/Modal.tsx
@@ -1,4 +1,5 @@
 import * as Dialog from "@radix-ui/react-dialog";
+
 import {
   FilterBody,
   FilterBodyInner,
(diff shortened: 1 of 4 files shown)


bmquinn (approved):
The query looks great in my dev env. It's really awesome seeing the hybrid query results coming in while the streaming response is being generated! 😄

PR #328, opened 2024-05-16
Prompt user login when ai chat query param is in the url

adamjarling:
## What does this do?

Prompt a user to login when `ai=true` query param exists in the url, and a user is not logged in.  This mainly targets new page loads or when sharing a deep link with someone not logged into NU.

image

### Option 1
1. Log out of the app, and then navigate to the search page and append `search?ai=true` to the end of the url (ie. https://devbox.library.northwestern.edu:3000/search?ai=true). 
2. Refresh the page and notice the Login alert dialog pops up

### Option 2
1. Log into the app, and enter a search term, maybe with facet/filters as well
2. Copy this URL
3. Open a new browser (make sure you're not logged in), and paste the URL into the browser address bar
4. Notice the Login alert dialog pops up

### DIFF
--- a/components/Search/GenerativeAIToggle.test.tsx
+++ b/components/Search/GenerativeAIToggle.test.tsx
@@ -84,24 +84,17 @@ describe("GenerativeAIToggle", () => {
       )
     );
 
-    const checkbox = screen.getByRole("checkbox");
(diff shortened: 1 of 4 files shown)


mathewjordan (approved):
Works like a dream.

PR #330, opened 2024-05-17
Split chat responses and legacy responses on `/search` screen

mathewjordan:
This work creates tabbing on the Search screen allowing a user to toggle between the streamed AI response and the legacy search results.

image

image

A few things:

- Tabbing is controlled at the search context level to better maintain control over the active view. 
- Updating search input should always reset the tab to the `stream` view if AI response is toggled on
- Updating facets on the results screen will should not reset the tab
- An actions bar will render below the streamed response upon completion, giving the user avenues to view the results or ask another question


adamjarling (approved):
Works great!  Looks awesome... let's get this in. 👍

PR #331, opened 2024-05-21
Facets filter button becomes fixed only after scrolling past chat response UI elements

adamjarling:
## What does this do?
Removes the Facets Filter button from relying upon the Fixed Search bar to trigger it's own fixed position in the center of the screen.  Instead, there's now a scroll event listener, and if the Filter element gets within a range of the top of the viewport (hardcoded for now to 70px), this calculation will trigger the Filter element's position on the screen.

image

image

## How to test
Perform a Chat AI search, and notice the placement of the Filter button when scrolling up and down the page.


mathewjordan (approved):
Nice, I'm falling in love with using these refs where possible.

PR #333, opened 2024-05-31
Replace iiif.stack.rdc.library with iiif.dc.library

mbklein:
The old IIIF base URL `https://iiif.stack.rdc.library.northwestern.edu/` is deprecated. Switching everything to `https://iiif.dc.library.northwestern.edu/`.


adamjarling (approved):
Looks good.

PR #335, opened 2024-06-13
Support Github flavored markdown with new NPM useMarkdown hook

adamjarling:
## What does this do?
Swaps out DC's custom `useMarkdown` hook with `@nulib/use-markdown`.

## How to test
- Verify search result text renders appropriate HTML from markdown
- Verify a trailing blinking cursor is present when the chat response is actively streaming and not finished


mathewjordan (approved):
💯 Awesome, glad to see this and that it still works well with the animated cursor.

Can we wait to merge after #330?

PR #336, opened 2024-06-17
Add placeholder texts to AI search

adamjarling:
## What does this do?
- Changes Search `input` element to `textarea` element
- `textarea` placeholder text changes based on whether AI chat is engaged
- Put in some language on AI results when no prompt has been entered

image

@davidschober @vlsrobinson @mathewjordan  Let me know about the placeholder text?

## How to test
- Verify placeholder texts
- Try copying or writing a lot of text into the search input to validate its a `textarea` and will wrap vertically.


mathewjordan (approved):
I like it. For the textarea, I think we need one more issue to flesh it out UX. 

- Detect textarea scrollHeight to auto grow the height of the area, so that as the user types and the textarea is in focus
- When user unfocuses, peel back height to minheight and add `...` text overflow to indicate a long wrapping string
- Remove fixed height of the header to allow for fluid growth by use of minheight where height has been set
- Refine form buttons so that they don't grow in height as well
- Test for fixed height
- Test responsively

mathewjordan:
Created https://github.com/nulib/repodev_planning_and_docs/issues/5045

vlsrobinson:
Should we prompt the user to ask a question, instead of 'try searching for..' maybe something like "try asking a question like, who are some musicians who played at the Berkeley Folk Music Festival?"

adamjarling:
> Should we prompt the user to ask a question, instead of 'try searching for..' maybe something like "try asking a question like, who are some musicians who played at the Berkeley Folk Music Festival?"

Sure that sounds like a good idea @vlsrobinson .  One UI consideration (now that this is a `textarea` element, is that the placeholder text needs to fit on one line, and avoid spilling onto a second line.  Roughly ~55 characters of space available.

adamjarling:
@vlsrobinson Because there are some new UI considerations to code for, I'm going to merge this now, and then we can flesh out the text at anytime.

PR #339, opened 2024-06-21
Update hybrid search pipeline.

mathewjordan:
I'm unsure if I am doing this correctly. The NEXT_PUBLIC_OPENSEARCH_PIPELINE was always `undefined` and I'm not sure what it should have been leaning on for me so I have rewritten this a bit.

```tsx
if (!!body?.query?.hybrid) {
  requestUrl.searchParams.append(
    "search_pipeline",
    "dc-v2-work-pipeline"
  );
}
```

Seems to do the trick... is there a reason we would want this to be an environmental variable?

image

### DIFF
--- a/pages/search.tsx
+++ b/pages/search.tsx
@@ -96,7 +96,6 @@ const SearchPage: NextPage = () => {
         const { page, q } = router.query;
         const urlFacets = parseUrlFacets(router.query);
         const requestUrl = new URL(DC_API_SEARCH_URL);
-        const pipeline = process.env.NEXT_PUBLIC_OPENSEARCH_PIPELINE;
 
         // If there is a "similar" facet, get the work and set the state
         if (urlFacets?.similar) {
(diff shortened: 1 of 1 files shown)


mbklein (approved):
👍🏻 The pipeline will fail (with an error) if applied to a non-hybrid query, so this is the correct and necessary way to do it. I thought of something else last night, but it's an improvement for another day.

adamjarling:
@mathewjordan @mbklein The pipeline variable should be pulling through ok.  I updated this a few weeks back in https://github.com/nulib/miscellany, and I thought I posted in Slack for folks to pull the latest `miscellany`, but if I forgot sorry!

https://github.com/nulib/miscellany/blob/479111a2f17ed45fe2e975e256a0d61a0c28ebee/secrets/dev_environment.sh#L36

mbklein:
@adamjarling Yeah, configuring it using a variable is a good plan, but it's also important _only_ to use it for hybrid queries. It'll break regular ones.

adamjarling:
@mbklein Yep, that should be how's it wired up, to only send `?search_pipeline` if doing hybrid (AI) querying.

mathewjordan:
Oh, well, we need to revert things. Ahhh.

PR #343, opened 2024-06-25
Move Gen AI toggle tab state to localStorage

adamjarling:
## What does this do?
- Defaults the active tab on Search page to show Chat streaming response when a user enters the Search route with the Gen AI checkbox checked.
- Moves Gen AI checkbox state from URL query params, to `localStorage`
- Removes active tab state from Search Context
- Attempts to handle faceting/filtering UX when Gen AI checkbox state is active

This PR moves the ball down the field, however there is one last piece which I'm not sure of the proper direction?

When a user is search w/ Gen AI, switches to the View Results tab, actively filters the results, and then clicks on the "AI Response" tab, the AI stream creates a new websocket request and doesn't render the cached result.   Is this an expected outcome or no?  Does faceting or filtering items change the nature of the original AI response, I guess is the question.

## How to test
Basically just click around the app, searching, toggling the Gen AI check box on off at various places and times.  Try logging in / out, see how the UX responds.  Anything which seems out of the ordinary should be noted and addressed.

### DIFF
--- a/components/Chat/Chat.test.tsx
+++ b/components/Chat/Chat.test.tsx
@@ -50,8 +50,7 @@ jest.mock("@/hooks/useChatSocket");
 }));
 
 describe("Chat component", () => {
-  it("renders default ai response placeholder text", () => {
-    mockRouter.setCurrentUrl("/search?ai=true");
+  it("renders default placeholder text when no search term is present", () => {
     render(<Chat />);
 
     const wrapper = screen.getByText(
@@ -64,7 +63,7 @@ describe("Chat component", () => {
   });
 
   it("renders the chat response component when search term is present", () => {
-    mockRouter.setCurrentUrl("/search?q=tell+me+about+boats&ai=true");
+    mockRouter.setCurrentUrl("/search?q=tell+me+about+boats");
 
     render(
       <SearchProvider>
@@ -94,7 +93,7 @@ describe("Chat component", () => {
       sendMessage: mockMessage,
     }));
 
-    mockRouter.setCurrentUrl("/search?q=boats&ai=true");
+    mockRouter.setCurrentUrl("/search?q=boats");
 
     render(
       <SearchProvider>
@@ -112,7 +111,7 @@ describe("Chat component", () => {
   });
 
   it("doesn't send a websocket message if the search term is empty", () => {
-    mockRouter.setCurrentUrl("/search?ai=true");
+    mockRouter.setCurrentUrl("/search");
     render(
       <SearchProvider>
         <Chat />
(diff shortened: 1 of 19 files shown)


mathewjordan (changes requested):
This works great aside from one point of complexity. What should the user expect when the do the following?

1. Check _Use Generative AI_
2. Do their query/question, "Music"
3. Click _View `n` Results_ tab
4. Select any facet (Image)

Right now, this will select the facet however it will also send them back to the AI Response tab. I think this is where we keep falling back an forth. This bit of complexity makes it where we can't necessarily rely on the Checkbox solely for deciding the current tab -- but some other current state knowledge is necessary as well.

adamjarling:
> This works great aside from one point of complexity. What should the user expect when the do the following?
> 
> 1. Check _Use Generative AI_
> 2. Do their query/question, "Music"
> 3. Click _View `n` Results_ tab
> 4. Select any facet (Image)
> 
> Right now, this will select the facet however it will also send them back to the AI Response tab. I think this is where we keep falling back an forth. This bit of complexity makes it where we can't necessarily rely on the Checkbox solely for deciding the current tab -- but some other current state knowledge is necessary as well.

@mathewjordan I've been really trying to avoid suggested browser `localStorage` to track whether a user has opted into Gen AI, but your point above aligns with some other scenarios which might be pointing in that direction.  We can't really store it in top level Context (page refresh wipes it), seems clunky to force it into a URL param on every route...  Thoughts?  I can do a quick test and gauge how big a refactor this might be.

mathewjordan:
> > This works great aside from one point of complexity. What should the user expect when the do the following?
> > 
> > 1. Check _Use Generative AI_
> > 2. Do their query/question, "Music"
> > 3. Click _View `n` Results_ tab
> > 4. Select any facet (Image)
> > 
> > Right now, this will select the facet however it will also send them back to the AI Response tab. I think this is where we keep falling back an forth. This bit of complexity makes it where we can't necessarily rely on the Checkbox solely for deciding the current tab -- but some other current state knowledge is necessary as well.
> 
> @mathewjordan I've been really trying to avoid suggested browser `localStorage` to track whether a user has opted into Gen AI, but your point above aligns with some other scenarios which might be pointing in that direction. We can't really store it in top level Context (page refresh wipes it), seems clunky to force it into a URL param on every route... Thoughts? I can do a quick test and gauge how big a refactor this might be.

I'm all for it if it simplifies handling.

mathewjordan (commented):
I'm not sure if I'm missing something. I see the `ai` key in local storage now, however it doesn't seem to be defaulting to the AI Response as expected. Maybe we can check it out together?

image

mathewjordan (approved):
💯

PR #344, opened 2024-06-26
Block sending empty messages to chat endpoint

adamjarling:
## What does this do?
Handles a scenario where empty search values were being passed to the endpoint (causing issues in the API).  The `Chat` component now guards against this by checking for a search term value.

## How to test
1. Open the Network tab of your browser and navigate to the Websocket (WS) tab to inspect traffic.
2. Try searching with Gen AI enabled with both empty and non-empty search terms. 
3. Notice in the websocket logs, the `question` property will never be an empty string.

image

For developers, you can also do a quick console log, or put a breakpoint in this function to check as well:

https://github.com/nulib/dc-nextjs/blob/a6ddf55afd3e0815fe7b31e88c2305eaf5524c54/hooks/useChatSocket.ts#L63-L67


mathewjordan (approved):
👍

PR #346, opened 2024-06-26
Refine TextArea UX, overall header styling. 

mathewjordan:
This work refines the Search component and textarea UX. Now when a user begins to type, the `<textarea>` element will grow in by calculating the `scrollHeight` of the element and applying this height as a style. If the user unfocuses, the height should transition back to the default height (appx 50px) with a `text-overflow` rendering an ellipsis for overflowing textual content. 

To make this blend well with our current layout, I opted to simplify the background colors, making the previous background of purple now transparent and showing white. This should elevate the actionable elements in the header to the foreground.

Along with this, I also made some other stylistic updates regarding overall legibility of text in the `html` document, sharpening and modernizing things. Along with this, I also updated the styling of the AI Response / View `n` Results tab.

image

### Other Notes

- I needed to refine some `Search` components tests after making modifications. Mainily, I removed the need for us to know `isSearchActive` on the Generative AI toggle which was only used for margins. These updates should simplify things there.

- I needed to mock the ResizeObserver as Radix was screaming about this for some reason (see jest.setup)

### DIFF
--- a/components/Chat/Response/Response.styled.tsx
+++ b/components/Chat/Response/Response.styled.tsx
@@ -20,7 +20,6 @@ const StyledResponse = styled("section", {
 });
 
 const StyledResponseAside = styled("aside", {
-  // background: "linear-gradient(7deg, $white 0%, $gray6 100%)",
   width: "38.2%",
   flexShrink: 0,
   borderRadius: "inherit",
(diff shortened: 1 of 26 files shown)


adamjarling (approved):
Great work here, looks so much better, cleaner.  👍

PR #348, opened 2024-07-02
Update chat endpoint route

kdid:
Updated in https://github.com/nulib/dc-api-v2/pull/226 from `chat-endpoint` to `chat/endpoint`


bmquinn (approved):

mathewjordan (approved):
@adamjarling noted for his in progress PR but can't hurt to handle it here.

adamjarling:
Yep, I made the change too, but let's get it in quick 👍

adamjarling (approved):

kdid:
Actually @mathewjordan @adamjarling - I just realized you use production so we'll need to time the merge with that

PR #349, opened 2024-07-15
Add GTM dataLayer data to Shared and 404/403 pages

adamjarling:
## What does this do?
Adds Google GTM `dataLayer` data to Shared Work pages (`/shared/ID`), and `404` and `403` pages.

### Shared routes
1. Open up Meadow staging, and share a Work
2. Open DC in a dev environment which uses staging data, and copy/paste the `/shared/ID_HERE` string of the route do your local dev instance URL.
3. Open NUL GTM account in a browser, and click the "Preview" option.
4. Paste in your shared link URL
5. Validate DataLayer info is pulling through

### 403 / 404 routes
Same as above, put preview the "bad" routes, and notice that `dataLayer` data `pageTitle` will contain either "403 page" or "404 page"

### DIFF
--- a/lib/dc-api.ts
+++ b/lib/dc-api.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError } from "axios";
+import axios, { AxiosError, RawAxiosRequestHeaders } from "axios";
 
 import type { ApiSearchRequestBody } from "@/types/api/request";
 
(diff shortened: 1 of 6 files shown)


mathewjordan:
This method seems to work only for `Institution` items. For Private works, the network request may need to be client-side for cookies to be sent through...

- If wanting server-side, investigate Next cookies package?
- Or push client-side strategy again

mathewjordan (approved):

PR #350, opened 2024-07-18
Follow up Playwright tests clean up

adamjarling:
## What does this do?
- Adjusts `package.json` Playwright scripts to use a similar naming convention
- Adds `Prettier` check to CI Unit Test workflow
- Adds caching to the CI Unit Test workflow for improved performance
- Removes Cypress deps which re-appeared after a merge.


mathewjordan:
🚀

mathewjordan (approved):
Works great. I like the rename.

PR #355, opened 2024-07-26
Persist results tab if paging or facets are adjusted

mathewjordan:
## What does this do?

This adjusts how we are setting the active tab on search results page.

- Makes `results` the "default"
- Uses the router.query `q` to adjust tabs if new search sent through


adamjarling (approved):
Works like a charm

PR #357, opened 2024-07-29
Adjust flex to make textarea responsive on smaller viewports. 

mathewjordan:
## What does this do?

This applies a simple CSS flexbox fix to make our textera/form responsive at smaller viewports, wrapping the AI toggle and search submit to their own line.

image


adamjarling (approved):
🚀

PR #358, opened 2024-07-31
Add current build command config to Terraform

kdid:
- We don't need prettier on the actual build.
- This was already present in the Amplify console, just moving to Terraform so it doesn't get overwritten


mathewjordan (approved):
👍

PR #363, opened 2024-08-15
Add "View `x` results" callback.

mathewjordan:
## What does this do?

This adds a small callback on the **View `x` results** button that renders below a completed streamed AI response. The callback simply switches the active tab to `results`.

## How can we review?

- Navigate to preview branch at https://5159-view-results-toggle.dc.rdc-staging.library.northwestern.edu/
- Complete an AI chat search
- After streaming completes, click View `x` results below the response
- Tab should toggle to hybrid results

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -18,7 +18,13 @@ import { prepareQuestion } from "@/lib/chat-helpers";
 import useChatSocket from "@/hooks/useChatSocket";
 import useQueryParams from "@/hooks/useQueryParams";
 
-const Chat = ({ totalResults }: { totalResults?: number }) => {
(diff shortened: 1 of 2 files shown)


mbklein:
I can't execute an AI search on that site because of the login restriction. SSO/API logins only work on sites that end with `northwestern.edu` so using an `amplifyapp.com` address won't work.

mathewjordan:
> I can't execute an AI search on that site because of the login restriction. SSO/API logins only work on sites that end with `northwestern.edu` so using an `amplifyapp.com` address won't work.

This should be good now using the updated preview URL -- Thanks @kdid

kdid (approved):
👍

PR #364, opened 2024-08-16
Use next hooks to avoid renrender with previous query.

mathewjordan:
## What does this do?

This fixes an issue where the query params of the rendering page are delayed in the re-rendering due to a hiccup in a `useEffect()`.  updates our `useQueryParams()` hook to no longer utilize the `React.useEffect()`. In doing so, we won't see as many re-renders when the hook is called on components where it is used. 

## How should we review?

This isn't the most obvious thing to review as it is literally milliseconds between renders. But the best way is to:

 - Spin up DC or visit https://preview-5154.dc.rdc-staging.library.northwestern.edu/
 - Set to `Use Generative AI` and login
 - Do any search, let it complete
 - Do another search, question rendered on screen should immediately be replaced\
 

As this updates a hook used throughout DC, we should review legacy search:

 - Complete a non-AI search
 - Play with facets
 - Things should be 💯

### DIFF
--- a/hooks/useQueryParams.ts
+++ b/hooks/useQueryParams.ts
@@ -1,22 +1,12 @@
-import React from "react";
-import { UrlFacets } from "@/types/context/filter-context";
 import { parseUrlFacets } from "@/lib/utils/facet-helpers";
 import { useRouter } from "next/router";
 
 export default function useQueryParams() {
-  const { isReady, query } = useRouter();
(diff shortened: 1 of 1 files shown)


kdid (approved):
Looks 👍

kdid (changes requested):
I'm noticing some things now with facets/filters I'm not sure are issues or not. Lets walk through this tomorrow before merge?

kdid (approved):

PR #366, opened 2024-08-23
Remove number counts from Chat hybrid response results.

mathewjordan:
## What does this do?

This brushstroke sweeps away the result counts from the UI where rendered on chat responses.

 - View More Results buttons (tabs and response footer actions)
 - View Results modal submit button
 - Pagination results counts in footer (for 40+ hits)

Note, this does not include individual agg accounts for bucket items on the filter facets modal.

## How should we review?

- Go to https://preview-5175.dc.rdc-staging.library.northwestern.edu or branch in DC
- Complete any AI search
- Counts should be excluded for:
  + View More Results buttons in tab and footer actions should say "View More Results"
  + View Results Modal submit button should say, "View Results"
  + Pagination result counts should not be displayd

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -151,10 +151,10 @@ const Chat = ({
           <Container>
             <StyledResponseActions>
               <Button isPrimary isLowercase onClick={viewResultsCallback}>
-                View {pluralize("Result", totalResults || 0)}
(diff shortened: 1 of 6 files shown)


kdid (approved):
👍

bmquinn (approved):

PR #367, opened 2024-08-26
Initiate new query on isAi toggle.

mathewjordan:
## What does this do?

This minor fix makes it so that a new query is initiated on toggle of the _Use Generative AI_ option in the search component. Previously, this only fired if the query string itself changed and was submitted. 

To make sure work, we adjust the useEffect hook on the search page to watch for any changes in `isAi` state here by adding it as a dependency. 
https://github.com/nulib/dc-nextjs/blob/88887a4db113408ab628a5f209ad90427d611c2d/pages/search.tsx#L146

And then we also reset the query response data to wipe previous data results.
https://github.com/nulib/dc-nextjs/blob/88887a4db113408ab628a5f209ad90427d611c2d/pages/search.tsx#L92-L93

Additionally, I have abstracted out the `search.tsx` screen further, creating the `components/Search/Results.tsx` file. This should give us better control over rendering of results in the future.

## How should we review?

- Go to https://preview-5164.dc.rdc-staging.library.northwestern.edu/ or this branch in your dev DC
- Complete a search (legacy), "Africa" or "Denver"
- See the results... note the first few items
- Select the _Use Generative AI_ toggle
- Screen should update to show Chat response with "See More Results" as a spinning loading SVG.
- See more results... first few items should be different
- Untoggle _Use Generative AI_
- Results should clear, new results should render

### DIFF
--- a/components/Search/Results.tsx
+++ b/components/Search/Results.tsx
@@ -1,0 +1,54 @@
+import {
+  NoResultsMessage,
+  ResultsMessage,
+  ResultsWrapper,
+} from "@/components/Search/Search.styled";
+
+import Grid from "@/components/Grid/Grid";
+import PaginationAltCounts from "@/components/Search/PaginationAltCounts";
+import { SearchResultsState } from "@/types/components/search";
+import { pluralize } from "@/lib/utils/count-helpers";
+import useGenerativeAISearchToggle from "@/hooks/useGenerativeAISearchToggle";
+
+const SearchResults: React.FC<SearchResultsState> = ({
+  data,
+  error,
+  loading,
+}) => {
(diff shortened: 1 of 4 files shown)


kdid:
@mathewjordan - I think your steps work until I change the text in the search box but then I think it's not firing with the updated text when I toggle until I click the search button. Should it?

mathewjordan:
> @mathewjordan - I think your steps work until I change the text in the search box but then I think it's not firing with the updated text when I toggle until I click the search button. Should it?

Yeah, I suppose that makes sense. I've added functionality that submits the form if AI toggling occurs.

kdid (approved):

PR #380, opened 2024-09-03
Consolidate collection jump to functionality in search component.

mathewjordan:
## What does this do?

This work integrates more fully the JumpTo component intended for our Collections page within our Search.tsx component. Previously we had broken out a seperate search component specifically for the collection pages. With the advent of the Use Generative AI toggle, this different strategy for collection pages left the search component there lacking our nice updates.

We now determine if we are on a collection route and then use this boolean to append the JumpTo component below the search component.

image

## How do we review?

 - Go to 
https://5166-preview.dc.rdc-staging.library.northwestern.edu or your dev DC
 - Go to any Collection, ex: **16th-Early 20th Century Maps of Africa**
 - Your should not see the JumpTo component
 - Begin typing, "Sudan"
 - JumpTo listbox should expand below the search component
 - Links JumpTo listbox options allow for searching "In This Collection" or "All Digital Collections"
 - "In This Collection" option will navigate the user the `/search` route with the Collection faceted and the query string matching, ex "Sudan"

Notes:

 - This should work with Use Generative AI toggled either way
 - This should be fully responsive and collapse appropriately in small viewports
 - The JumpTo options should not be present on other routes, include the Browse Collections `/collections` page

### DIFF
--- a/components/Header/Header.styled.ts
+++ b/components/Header/Header.styled.ts
@@ -92,6 +92,7 @@ const Primary = styled("div", {
       fontFamily: "$northwesternSansRegular",
       display: "flex",
       height: "$gr5",
+      marginLeft: "$gr4",
 
       a: {
         color: "$purple",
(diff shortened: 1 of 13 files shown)


mbklein (approved):
Just as advertised!

PR #391, opened 2024-12-04
Add AI `expires` value to local storage.

mathewjordan:
## What does this do?

This work extends the `ai` Local Storage key to be stringified JSON object with values of:

 - `enabled`
 - `expires`

On toggle (if signed in) of Use Generative AI, the `expires` value will be set as the current time + 60 minutes. On each hard refresh of site this expires value is checked against the current time. If current time is greater than that of the stored `expires` value, the local storage will be reset and `enabled` will be `false`. This `false` value will make it so that the Sign In dialog is not spammed.

image

Additionally, a small accessibility issue was addressed on the Sign In dialog due to a missing title.

## How should we review?

- Go to https://preview-5300-expire-ai-local-storage.dc.rdc-staging.library.northwestern.edu/ or spin up this branch in your dev environment
- Be on the homepage make sure you are signed out and Use Generative AI toggle is false
- In your browser, use your inspector to find Local Storage and not the `ai` key has a value of `{"enabled":"false"}`
- Click the Use Generative AI toggle and Sign In dialog shows
- On redirect to homepage, now note the the `ai` key has a value of `{"enabled":"true","expires":1733422722163}` where expires is the number set for one hour from the moment of toggle

Now test toggling again...

- Watch the inspector local storage value as your toggle Use Generative AI off/on
- Expires should reset each time to a new value
- Enabled should reset as expected
- Leave it on

Now to sign out to mimic the NetID logout/expiration that occurs every so often

-  With Use Generative AI toggled on, **copy** the AI key value `{"enabled":"true","expires":1733422722163}`
- Sign out and get yourself back to the homepage
- No dialog should show

But now, let's fake it as if we had left for the day and come back the next morning.

 - In your Local Storage, up the `ai` key value to be what you had copied earlier
 - Refresh and see the dialog
 - Now modify the expires number to some value in the obvious past by removing notching down the digits, ex: `{"enabled":"true","expires":1033422722163}`
 - Do a hard refresh of the page
 - The `ai` value should fully reset to be `{"enabled":"false"}` and the dialog should not show
 
And yeah, that's it. 🎉

### DIFF
--- a/components/Header/Super.tsx
+++ b/components/Header/Super.tsx
@@ -15,7 +15,9 @@ import { NavResponsiveOnly } from "@/components/Nav/Nav.styled";
 import { NorthwesternWordmark } from "@/components/Shared/SVG/Northwestern";
 import React from "react";
 import { UserContext } from "@/context/user-context";
+import { defaultAIState } from "@/hooks/useGenerativeAISearchToggle";
 import useLocalStorage from "@/hooks/useLocalStorage";
+import { useRouter } from "next/router";
 
 const nav = [
   {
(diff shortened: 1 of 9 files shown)


bmquinn (approved):
The testing steps worked as described for me in the preview branch, looks awesome 👍

mathewjordan:
Thanks @bmquinn!

PR #393, opened 2024-12-12
Show leading zeros on minutes in shared link expiration display

kdid:
- Shared links are currently displaying minutes in the expiration without leading zeroes.
- This update will now display, for example: `1:06pm` instead of `1:6pm`

### Steps to Test

- create a shared link in Meadow when the minute is a single digit
- view the shared link in DC and verify correct display of expiration time

### DIFF
--- a/lib/utils/date-helpers.test.ts
+++ b/lib/utils/date-helpers.test.ts
@@ -1,0 +1,73 @@
+import { formatDate, formatDateLong } from "./date-helpers";
+
+describe("formatDate", () => {
+  test("should format date correctly", () => {
+    const result = formatDate("2023-11-14T13:14:00Z");
+    expect(result).toBe("2023-10-2"); // Note: getMonth() is zero-based, getDay() returns day of the week
+  });
+
+  test("should handle invalid date", () => {
+    const result = formatDate("invalid-date");
+    expect(result).toBe("NaN-NaN-NaN");
+  });
+});
+
+describe("formatDateLong", () => {
+  it("should return a properly formatted date string for a valid date input", () => {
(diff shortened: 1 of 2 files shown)


lib/utils/date-helpers.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  test("should handle invalid date", () => {
+    const result = formatDate("invalid-date");
+    expect(result).toBe("NaN-NaN-NaN");
+  });
+});
+
+describe("formatDateLong", () => {
+  test("should format date correctly in long format", () => {
+    const result = formatDateLong("2023-11-14T13:14:00Z");
+    expect(result).toBe("November 14, 2023, 1:14pm");
```
mathewjordan, line 18: I'm getting:

```
formatDateLong › should format date correctly in long format

    expect(received).toBe(expected) // Object.is equality

    Expected: "November 14, 2023, 1:14pm"
    Received: "November 14, 2023, **8:14am"**
```

I think this might be related timezones and the `Z` in the date. Not sure how to best proceed here but I'm happy to sit with you to figure out your local tests.

mathewjordan (changes requested):
I think timezone might be an issue now (or always was). See my comment below:

lib/utils/date-helpers.test.ts
mathewjordan: image

kdid: @mathewjordan - updated to account for local time zones. The test is a little wonky but hopefully good enough.

kdid:
@mathewjordan - this is ready for a re-review.

mathewjordan (approved):
👍

PR #395, opened 2024-12-17
Preview/types 5322 conversations UI

charlesLoder:
Update chat types for better safety.

The build and playwright tests pass, but I was unable to test locally myself (having some install error with npm and getting env set up)


mathewjordan (approved):
🎸

PR #398, opened 2025-01-28
Handle conversations chat UI.

mathewjordan:
## What does this do?

This work makes major changes to the Chat response UI allowing for threaded conversations that are held in context through a `conversationRef`. A few updates have occurred to accommodate this change and iterate towards more complex multi question/answer responses.

1. useChatSocket is now handled at lower level component. This allows each question/answer to interact independently when receiving messages. Although indepdenent of each when receiving/sending messages, only one connection should be made at page level. This connection should sustain unless it times out.
 
https://github.com/nulib/dc-nextjs/blob/29f2c38eb544fb1576b492ec6fd74a454551186c/components/Chat/Response/Response.tsx#L27

2. The feedback form is now tied to each response. Note that the form UI is in place, however, the submission of this form and it's question/responses is not yet functional. Functionality for this will be handled in a new issue.

3. The more results table for hybrid search is hidden currently. This will be handled in a new issue and may require refinement before resolving.

4. We are not currently sending questions/answers to a higher level search context as we did previously. I have not yet determined if we need this search context any longer and what shape this should be. Solutions for (2) and (3) above should determine how we handle this. Tests related to mocking this context have temporally been removed.

## Review steps

This can be reviewed in amplify at https://preview-5354-conversations-ui.dc.rdc-staging.library.northwestern.edu/ or if run in a dev environment. If running in dev, be sure  to set `NEXT_PUBLIC_DCAPI_ENDPOINT`  to `https://dcapi-prototype.rdc-staging.library.northwestern.edu/api/v2`

image

### DIFF
--- a/components/Chat/Chat.test.tsx
+++ b/components/Chat/Chat.test.tsx
@@ -80,11 +80,6 @@ describe("Chat component", () => {
     const dataProps = el.getAttribute("data-props");
     const dataPropsObj = JSON.parse(dataProps!);
     expect(dataPropsObj.question).toEqual("tell me about boats");
-    expect(dataPropsObj.isStreamingComplete).toEqual(false);
-    expect(dataPropsObj.message).toEqual({
-      answer: "fake-answer-1",
-      end: "stop",
-    });
     expect(typeof dataPropsObj.conversationRef).toBe("string");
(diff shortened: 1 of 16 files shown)


kdid (approved):
👍

charlesLoder (approved):
The new icons are *chefs' kiss*

PR #405, opened 2025-03-03
Gracefully handle window scroll for more results

charlesLoder:
## Summary of changes

- Updates the "view more results" page size from 10 to 40
- Dynamically handles the height of the content when switching between chat and "View more results"
- When transitioning from "View more results" back to chat, window scrolls back to the "View more results" that was clicked

## Testing


pages/search.tsx
```diff
@@ -237,27 +244,18 @@ const SearchPage: NextPage = () => {
...
-              style={{
-                position: "relative",
-                overflow: "hidden",
-                marginTop: "-30px",
-              }}
-            >
+            <StyledTabsContent value="stream">
               <div
                 style={{
-                  position: "relative",
```
mathewjordan, line 250: I the removal of relative somewhere may have knocked loose the floating **Filter** button. This might be the culprit.

mathewjordan (changes requested):
This looks really great. Just one thing related to the Filter button.

image

pages/search.tsx
charlesLoder: This was tricky! Because the `Panel` has a `transform` applied to it, the filter button was in a new stacking context, and being positioned only according to the new containing block, see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/position)

> Fixed positioning is similar to absolute positioning, with the exception that the element's [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Containing_block) is the initial containing block established by the viewport, unless any ancestor has transform, perspective, or filter property set to something other than none (see [CSS Transforms Spec](https://www.w3.org/TR/css-transforms-1/#propdef-transform)), which then causes that ancestor to take the place of the elements [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Containing_block).

components/Search/Panel.tsx
```diff
@@ -72,7 +72,7 @@ const SearchPanel = () => {
         const requestUrl = new URL(DC_API_SEARCH_URL);
         const body: ApiSearchRequestBody = buildQuery(
           {
-            size: 40, // SEARCH_RESULTS_PER_PAGE
+            size: SEARCH_RESULTS_PER_PAGE,
```
mathewjordan, line 75: Good catch, didn't see that!

mathewjordan (changes requested):
Hey @charlesLoder I'm seeing some additional issues:

- The Filter button transitioning to the center no longer seems smooth. It previously would move from alignment left to the center using a CSS transition. There also seems to be an jump/flicker caused now when switching to center, something that was previously a bit more smooth. This might be related to an abrupt loss of a height in the container when the toggle takes place.

- There is now additional white space below the 40 search results if a conversation grows higher in height than results. I don't know if this was an issue yesterday but I don't recall seeing it. 

I can pair if it helps to work through this and if you need additional context to what I'm seeing.

mathewjordan:
Additionally, I'm now seeing the Filter button present on the conversation view.

PR #407, opened 2025-03-05
Update query aggregations to show correct results for facet counts on AI aided responses.

mathewjordan:
## What does this do?

This updates our queries by adding a `post_filter` on **query** and `filter` within **aggs**. These changes affected the response shape of `aggregations` and required some adjustments within components handling this data and types defining its shape.

#### `post_filter` added to query body

https://github.com/nulib/dc-nextjs/blob/7fa0c5356a1a355bd3a556bfd8cb0b55fece041d/lib/queries/aggs.ts#L82-L86

#### `filter` added to aggs

https://github.com/nulib/dc-nextjs/blob/7fa0c5356a1a355bd3a556bfd8cb0b55fece041d/lib/queries/aggs.ts#L58-L62

https://github.com/nulib/dc-nextjs/blob/7fa0c5356a1a355bd3a556bfd8cb0b55fece041d/lib/queries/aggs.ts#L82-L86

## How should we review?

1. Go to https://preview-5432-ai-aggs.dc.rdc-staging.library.northwestern.edu/search
2. Set **Use Generative AI**
3. Do a search that would bring up some fun results, ex: _wisconsin football_
4. Get that conversation rolling, and click on **View Results →** for the first search results interstitial
5. Click **Filter** and facet for subject that might cut the total results down a bit but not something extremely specific. If you went with _wisconsin football_, this might be **College Students**

image

6. After clicking, this facet being selected should filter out other facets that are exclusive of it.

image

7. Selecting another option should further drill in and make available options less abundant.

image

8. Toggling to another Facet, ex: **Genre** should also show only available options.

image

9. Click View Results should show results accurately.

image

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -101,7 +101,7 @@ const Chat = () => {
 };
 
 const StyledChat = styled("section", {
-  padding: "$gr6 0",
+  padding: "$gr5 0",
   position: "relative",
 });
 
--- a/components/Facets/Facet/GenericFacet.test.tsx
+++ b/components/Facets/Facet/GenericFacet.test.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from "@/test-utils";
+
 import GenericFacet from "./GenericFacet";
 import mockRouter from "next-router-mock";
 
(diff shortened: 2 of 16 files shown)


davidschober:
Approved. This works so well. AND it feels way more efficient.

davidschober (approved):

kdid (approved):

mbklein (approved):
This is awesome. :+1:

davidschober:
oh gosh this means that collection searching will just work! We do post_filter on collection ID. Sorry Just got excited again.

PR #411, opened 2025-03-24
Add docs to chat conversation

charlesLoder:
## Description

Adds the ability to add search results back into the conversation history as implemented in the [backend PR](https://github.com/nulib/dc-api-v2/pull/293)

## Summary of changes

- Adds `userDocs` to each `Turn`
- Adds `latestDocs` to the conversation history
- Users can optionally add filtered results to the chat history
- Users can remove results from history
- Users can ask about those docs specifically

## Example

Live: https://preview-5394-docs-to-convo.dc.rdc-staging.library.northwestern.edu/

Ask about a subject and go to "View results"

Screenshot 2025-03-24 at 4 15 48 PM

Filter results — will add a checkbox for "Add results to conversation". If checkbox is deselected, the user just returns to the conversation as normal

Screenshot 2025-03-24 at 4 16 18 PM

Click "Back to the conversation" and return to the chat with a new interstitial. At this point nothing has been submitted to the API. The user can remove the docs if they desire. Note that the new chat prompt is "Ask about these documents".

Screenshot 2025-03-24 at 4 16 39 PM

Ask a question about the selected documents:

Screenshot 2025-03-24 at 4 17 34 PM

## Notes

- There is a discrepancy where the `retrieve_documents` tool message doesn't seem to search for all the documents (or it isn't indicating that it is)

## Loose ends

Just jotting down things to come back to, no action needed now:
- docs in Feedback?
- naming conventions — docs vs results vs ???
- filter styling is still cut off

### DIFF
--- a/components/Chat/Response/InterstitialDocuments.tsx
+++ b/components/Chat/Response/InterstitialDocuments.tsx
@@ -1,0 +1,85 @@
+import { useEffect } from "react";
+import type { SearchContextStore } from "@/types/context/search-context";
+import {
+  StyledInterstitial,
+  StyledInterstitialWrapper,
+  StyledInterstitialIcon,
+  StyledInterstitialAction,
+} from "../Response/Interstitial.styled";
+import ResponseImages from "../Response/Images";
+import { IconClear, IconSparkles } from "@/components/Shared/SVG/Icons";
+import { useSearchState } from "@/context/search-context";
+
+interface InterstitialDocumentsProps {
+  documents: NonNullable<SearchContextStore["conversation"]["latestDocs"]>;
+  canRemove: boolean;
+}
+
+const InterstitialDocuments = ({
+  documents,
+  canRemove,
+}: InterstitialDocumentsProps) => {
+  const { searchState, searchDispatch } = useSearchState();
+
+  const {
+    panel: { open, interstitial },
+    conversation,
+  } = searchState;
+
+  useEffect(() => {
+    // if the panel is being opened or there is no interstitial, exit
+    if (open || !interstitial) return;
+
+    if ("latestdocs" !== interstitial) return;
+
+    const interstitialElement = document.getElementById(
+      `interstitial-latestdocs`,
+    );
+
+    if (!interstitialElement) return;
+
+    // because the search bar is absolutely positioned,
+    // grab its height to offset the scroll position
+    const searchBar = document.getElementById("dc-search");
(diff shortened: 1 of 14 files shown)


components/Chat/Response/InterstitialDocuments.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <>
+      <StyledInterstitialWrapper
+        id="interstitial-latestdocs"
+        data-testid="interstitial-latestdocs"
+      >
+        <StyledInterstitial>
+          <StyledInterstitialIcon>
+            <IconSparkles />
+          </StyledInterstitialIcon>
+          Selected documents ({documents.length}):
```
mathewjordan, line 73: I wonder if we can just leave off documents.length here? keep it vague?

components/Chat/Response/InterstitialDocuments.tsx
```diff
@@ -0,0 +1,85 @@
...
+      type: "updateConversation",
+      conversation: {
+        ...conversation,
+        latestDocs: undefined,
+      },
+    });
+  };
+  return (
+    <>
+      <StyledInterstitialWrapper
```
mathewjordan, line 65: Thoughts, perhaps this can be part of a new issue/iteration.

My perspective is that the wrapper isn't obviously part of the question because it resembles the AI Search Results interstitial.

- Can we wrap this in a color matching our Question bubble to make it more obvious that it is user created? 
- Maybe even NOT making it full width, 2/3 or 61.8%? Aligning right? And/or bringing it closer to the textarea, or somehow wrapper by the textarea.
- Smaller thumbnails? Not showing all, but a stack of them. Obviously this means we need a new component.

mathewjordan (approved):
I have a lot of thoughts but I don't think it fundametally gets in the way of the purpose of the issue. We can ask questions! We have context! Please take my ideas with a grain of salt, and maybe we can pair to experiment and refine early in the next cycle.

charlesLoder:
@mathewjordan see latest changes:

- removed the length in the `InterstitialDocuments` message
- removed the `Retrieving documents` interstitial
- add small thumbnails to the question when a user submit a question with docs

I didn't yet make the thumbnails in the `InterstitialDocuments` smaller.

I'm trying to keep the distinction between when a user navigates back to the chat and sees thumbnails (meaning they haven't actually asked about those docs yet) and between when they actually ask a question with docs attached.

It's still not there, but hopefully on the right track

PR #412, opened 2025-03-26
Add a reusable component for displaying user docs

charlesLoder:
## Description

Improves UI/UX by adding a reusable component for adding documents to the conversation context

## Summary of changes

- Creates a reusable "stack" component
- Adjusts app state by adding:
  - `conversation.context` which is context that has not been added directly into the turns of the conversation
  - `conversation.turns.context` which is context that has been added into the turn
- Minor styling adjustments

## Example

Ask a question and click on "View more results"

image

- By default the input is **not** checked
- With the input unchecked, navigating "Back to conversation" works as it has prior

Next, filter on something:

image

- when a filter is first chosen, the input is automatically checked
- there is a tooltip displaying the query and filters

Next, with the checkbox checked navigate with "Back to conversation"

image

- only the first 20 documents are passed into context
- the documents are added to the textbox, but with a dismiss button in case the user does not want to ask about the documents
- clicking dismiss removes them from the context

Next, asking a question with the docs in context shows them in question

image

- No tool call is displayed


mathewjordan:
I think it looks really great.

The one minor thing that has me is the Dismiss button.

Can we treat it closer to the Grey/Red "X' clears like other dismissables/close buttons?

image

image

charlesLoder:
@mathewjordan 
See update:
- there's no inset box shadow because there's too much visually happening with that component already
- updated the colors for consistency with the other component
- The padding on this one is different so that "X" icon is visible (with `padding: "$gr1"` it was just too small)

mathewjordan (approved):
Let's roll it in. 🔥

PR #158, opened 2023-01-12
Address Bloom related errors.

mathewjordan:
## What does this do?

This work brings in the updated Bloom IIIF component, addressing error handling for various resource type errors. Additionally, the `Next Page` slide is now hidden by targeting a `data-type` attribute on the swiper slide. Finally, a bug related to `&` character in strings of constructed IIIF Collections URIs that source Bloom sliders for series have been fixed.

- Update bloom dependency, hide next page,
- Address series IIIF collection construction.

## How to review?

1. Go to the Berkley collection
2. Click the **Collection Organization** tab
3. All sliders should render and function without error.

PR #414, opened 2025-04-07
Update escaping of double quotation marks and encoding

charlesLoder:
## Description

Updates handling of escaping of double quotation marks.

## Summary or changes

- escape double quotation marks inside of `query=series:"<TERM>"`
- encode the `collectionId` inside `<SliderIIIFWrapper />`
- escape double quotation marks in `iframe` string
- fix minor bug when trying to render the `iframe`

See additional annotations in comments


components/Collection/Tabs/Organization.tsx
```diff
@@ -34,17 +36,16 @@ const CollectionTabsOrganization: React.FC<CollectionTabsOrganizationProps> = ({
           const summary =
             doc_count > 1 ? `${doc_count} Items` : `${doc_count}  Item`;
 
-          const sanitizedKey = key.replace(/&/g, "%26");
-          const collectionId = `${DCAPI_ENDPOINT}/search?query=series:"${sanitizedKey}"&collectionLabel=${sanitizedKey}&collectionSummary=${summary}&as=iiif`;
+          const collectionId = `${DCAPI_ENDPOINT}/search?query=series:"${escapeDoubleQuotes(key)}"&collectionLabel=${key}&collectionSummary=${summary}&as=iiif`;
```
charlesLoder, line 39: This partially reverts changes made in 2d5fa5bf6c26aaa096a7b8b3a1cf39ff955e6447

There is no reason (I can see) to percent encode the `&` character separately because:
- when the `collectionId` is passed down through the `SliderWrapper` it is encoded there
- using `URL.searchParams` automatically does encoding

But perhaps it needs to be escaped within `query=series:"<TERM>"` for reasons unknown to me

pages/embedded-viewer/[manifestId].tsx
```diff
@@ -64,9 +64,7 @@ const EmbeddedViewerPage: NextPage<EmbeddedViewerPageProps> = ({ work }) => {
 export const getServerSideProps: GetServerSideProps = async (context) => {
   const mId = decodeURIComponent(context?.params?.manifestId as string);
 
-  const workId = mId
-    .replace(`${DCAPI_ENDPOINT}/works/`, "")
-    .replace("?as=iiif", "");
+  const workId = new URL(mId).pathname.split("/").pop()!;
```
Comment (charlesLoder) [inline pages/embedded-viewer/[manifestId].tsx:67]: When the `showTitle` option is set to `false`, the embedded viewer does not render:

```html
<iframe 
  src="https://dc.library.northwestern.edu/embedded-viewer/https%3A%2F%2Fapi.dc.library.northwestern.edu%2Fapi%2Fv2%2Fworks%2F6fcec77d-3c76-4bf4-a53b-626bb9c315f5%3Fas%3Diiif%26showTitle%3Dfalse"
  title="Opera Workshop - "Country Doctor"" 
  width="100%" 
  height="800">
 </iframe>
 ```

File: components/Collection/Tabs/Organization.tsx [hunk 3]
```diff
@@ -34,17 +36,24 @@ const CollectionTabsOrganization: React.FC<CollectionTabsOrganizationProps> = ({
...
           const title = entry.key;
           const indicator = pluralize("Item", entry.doc_count);
 
           const searchUrl = new URL("/search", DC_URL);
           searchUrl.searchParams.append("collection", collection.title);
-          searchUrl.searchParams.append("series", sanitizedKey);
+          searchUrl.searchParams.append("series", key);
 
-          if (isAI) appendHybridSearchParams(searchUrl, sanitizedKey);
+          if (isAI) appendHybridSearchParams(searchUrl, key);
```
charlesLoder, line 57: This partially reverts changes made in 2d5fa5bf6c26aaa096a7b8b3a1cf39ff955e6447 (and see related PR #158)

There is no reason (I can see) to percent encode the `&` character separately now because using `URL.searchParams` automatically does encoding.

But perhaps it needs to be encoded within `query=series:"<TERM>"` for reasons unknown to me, but the original change seems to be related to Bloom

components/Shared/RelatedItems.tsx
```diff
@@ -29,7 +29,7 @@ const RelatedItems: React.FC<RelatedItemsProps> = ({ collections, title }) => {
       <ErrorBoundary FallbackComponent={ErrorFallback}>
         {collections?.map((workSlider) => (
           <SliderWrapper
-            collectionId={workSlider.iiifCollectionId}
+            collectionId={new URL(workSlider.iiifCollectionId)}
```
mathewjordan, line 32: I love passing the `URL` constructor in, but it feels awkward to instantiate it inline. Can we do define this as a `const` outside of the return and then pass it?

mathewjordan (changes requested):
Looks great. One small note regarding the `URL()` constructor placement.

mathewjordan (approved):

components/Shared/RelatedItems.tsx
mathewjordan: Ah, I see why you did this now. Was not seeing that the `map`.

PR #417, opened 2025-04-11
Address timing issue related to initial question.

mathewjordan:
## What does this do?

This addresses a timing issue related to an initial question, where the question was only set in the search context if the user was already on the search page and refreshed. All logic for setting of the initial question and the conversation ref (a uuid) has now been moved to the Search component. Now the initial question is only set in the case of a form submission in the Search.tsx component.

Additionally, if the Generative AI checkbox is toggled true in this component, the conversation ref is reset, creating a new conversation. This flushing of the ref was necessary in cases of a user toggling back and forth between semantic conversations and legacy search.

## How should we review?

Go to https://preview-initial-conversation-question.dc.rdc-staging.library.northwestern.edu/

A) Homepage start

 - Search something, ex: `jim roberts`
 - Set AI Gen checkbox `true`
 - Submit
 - Question should render in purple bubble
 - Answer should be streaming in
 
B) Search page start

 - Go to search page first, see legacy results
 - Set AI Gen checkbox `true`
 - Should toggle to chat view
 - Ask a question
 - Question should render in purple bubble
 - Answer should be streaming in

C) Followup questions should function as expected
D) Search new question from Top search bar should reset conversation as expected

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -7,38 +7,18 @@ import Container from "@/components/Shared/Container";
 import { StyledUnsubmitted } from "./Response/Response.styled";
 import { styled } from "@/stitches.config";
 import { useSearchState } from "@/context/search-context";
-import { v4 as uuidv4 } from "uuid";
 
 const Chat = () => {
-  const { searchState, searchDispatch } = useSearchState();
-
-  const { conversation } = searchState;
-
-  const ref = conversation.ref;
-  /** get initial question  */
-  const initialQuestion = conversation.initialQuestion;
-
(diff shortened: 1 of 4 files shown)


charlesLoder (approved):
This feels a lot tighter

I haven't tested it, but in

https://github.com/nulib/dc-nextjs/blob/81dc170b7597b1a8ad70daf0f85b6b8ff40cae9f/types/context/search-context.ts#L25-L39

It looks like the `ref` prop doesn't have to be optional now, so checks like in Chat.tsx:46 could be removed too. Not necessary though

bmquinn (approved):
Testing out the deployed preview branch, don't notice any timing issues 👍

PR #418, opened 2025-04-18
Update local URLs to work with wildcard cert

mbklein:
- Change all instances of devbox.library to local.dev.rdc.library to work with wildcard dev cert


mathewjordan (approved):
🤩

PR #419, opened 2025-04-28
Add trailing slash to DC API endpoint used for base urls.

mathewjordan:
## What does this do?

Appends a trailing slash to the DCAPI_ENDPOINT constant to make it friendly for URL constructors. This retains `/api/v2/` pathname and appends `search`.

### DIFF
--- a/components/Collection/Tabs/Organization.tsx
+++ b/components/Collection/Tabs/Organization.tsx
@@ -25,6 +25,10 @@ const CollectionTabsOrganization: React.FC<CollectionTabsOrganizationProps> = ({
 
   const list = sortAggsByKey(series);
 
+  // creates new API endpoint string that has a trailing
+  // slash and is path friendly for a URL constructor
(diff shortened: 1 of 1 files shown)


charlesLoder (approved):
There are still some `console.log` statements if you want to get rid of them, but they don't hurt either. thanks for figuring out this obscure knowledge!

components/Collection/Tabs/Organization.tsx
```diff
@@ -55,6 +59,9 @@ const CollectionTabsOrganization: React.FC<CollectionTabsOrganizationProps> = ({
 
           if (isAI) appendHybridSearchParams(searchUrl, key);
 
+          console.log(`DCAPI_ENDPOINT`, DCAPI_ENDPOINT);
+          console.log(`collectionId`, collectionId.toString());
+
```
mathewjordan, line 64:
```suggestion
```

PR #421, opened 2025-05-07
Add link to system prompt

charlesLoder:
## Summary of changes

- Adds a link to the current system prompt (1) in the Gen AI tooltip (2) below the user input area

## Examples

Below the user input:
image

In the tooltip:

image

## Notes

- I added it a constant instead of a new component because it didn't feel important enough to be a component (original vibe coding) or have state or have children, but also if you prefer it be inside of `/components` I'm happy to change that easily.
- I let the parents control the styling so that it would match the tooltip styling better, but we can also keep it consistently styled wherever it is
- Also updated the feedback form to `visibility: "hidden"` instead of `opacity: 0` which improves keyboard navigation

### DIFF
--- a/lib/constants/common.tsx
+++ b/lib/constants/common.tsx
@@ -2,5 +2,19 @@ export const AI_DISCLAIMER = `This is a preview of Digital Collection's semantic
 export const AI_LOGIN_ALERT = `You must be logged in with a Northwestern NetID to use the Generative AI search feature.`;
 export const AI_SEARCH_UNSUBMITTED = `What can I help you find? Try searching for "john cage scrapbooks" or "who played at the Berkeley Folk Music Festival in 1965?"`;
 export const AI_TOGGLE_LABEL = "Use Generative AI";
+export const AI_SYS_PROMPT_MSG = () => (
+  <span className="ai-sys-prompt-msg">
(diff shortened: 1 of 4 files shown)


lib/constants/common.tsx
```diff
@@ -2,5 +2,19 @@ export const AI_DISCLAIMER = `This is a preview of Digital Collection's semantic
 export const AI_LOGIN_ALERT = `You must be logged in with a Northwestern NetID to use the Generative AI search feature.`;
 export const AI_SEARCH_UNSUBMITTED = `What can I help you find? Try searching for "john cage scrapbooks" or "who played at the Berkeley Folk Music Festival in 1965?"`;
 export const AI_TOGGLE_LABEL = "Use Generative AI";
+export const AI_SYS_PROMPT_MSG = () => (
+  <span className="ai-sys-prompt-msg">
+    Curious how this works? Click{" "}
+    <a
+      href="https://github.com/nulib/dc-api-v2/blob/main/chat/src/agent/search_agent.py#:~:text=DEFAULT_SYSTEM_MESSAGE"
+      target="_blank"
```
mathewjordan, line 10: 👍

mathewjordan (approved):
Looks great to me!

PR #423, opened 2025-05-13
Update collections page using infinite scroll

charlesLoder:
## Summary of changes

Improves the loading of the `/collections` page by implementing infinite scrolling.


mathewjordan (changes requested):
Some thoughts: 

1) This seems to break the filter/search feature as only results from the initially loaded sets of collection will be available. We may need to at least have basic metadata for all collections available or rethink how that search works. As a quick fix, I'd opt for the first.

2) The continuous scroll is great! However, I see it being a little _herky-jerky_ if the user scrolls at a moderately quick speed. I wonder if a size of `10` could help with that. From what I can tell, it could lessen to payload of each next page load by about 100kb (200kb --> 100kb).

charlesLoder:
@mathewjordan see the latest iteration taking a different approach

mathewjordan (changes requested):
This resolves the initial load really cleanly. I do note that if I load the page and immediately begin searching/filtering for "World" I will see a **"No results found for world."** message. This seems to be because the request hasn't been made yet to add those items to `entireCollectionsList`. Can we serverside/preload a full listing somehow? Let me know if this is complicating things further.

charlesLoder:
> Can we serverside/preload a full listing somehow? Let me know if this is complicating things further.

Yes and no. We can preload everything, but slows down the initial page response time b/c it's fetching all the collections and the HTML becomes pretty huge. I can try the `/search` api to return less data…let me see how that works

charlesLoder:
@mathewjordan see latest iteration. Trimmed down the server props size by using the `search` api so everything is loaded initially

mathewjordan:
:shipit: Ship it!

mathewjordan (approved):

PR #424, opened 2025-05-14
Adjust DC Next for new token scopes configuration

kdid:
### Summary

Along with this [DCAPI change](https://github.com/nulib/dc-api-v2/pull/314), the JWT now has `scopes` which define what the user can access. 

```
  read:Public
  read:Published
  read:Institution
  read:Private
  read:Unpublished
  chat
```

Additionally there has been an `isInstitution` claim added to distinguish between NUSSO and MagicLink logins

This work uses these scopes as much as possible and removes logic from front end.

### Specific Changes in this PR

- MagicLink users should behave as anonymous users except for chat access
  - see locks by thumbnails in search results for Institution works
  - See blurred bg image instead of Clover viewer for Institution works
  - Same Download and Share Tab restrictions as anonymous users
  - Respect existing Reading Room restrictions

### How to test

- Run dev app against staging (or local) API
- Check anonymous, Northwestern and MagicLink (& reading room. if possible) users for appropriate access to works across the site

### DIFF
--- a/components/Clover/ViewerWrapper.test.tsx
+++ b/components/Clover/ViewerWrapper.test.tsx
@@ -10,10 +10,14 @@ const userContextValue = {
     isReadingRoom: false,
     name: "Joan Doe",
     sub: "jdoe2399",
+    isInstitution: true,
+    scopes: ["read:Public", "read:Institution", "read:Published", "chat"],
+    primaryAffiliation: "staff",
+    provider: "nusso",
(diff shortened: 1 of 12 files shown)


mbklein (approved):
Looks great! Works great! 👍🏻

mathewjordan (changes requested):
Just some small types changes.

PR #425, opened 2025-05-14
Update auth to handle magic links

charlesLoder:
## Description

Refactors the UX for auth by allowing for Magic Links and routing all auth through a dialog (instead of just when toggling the gen ai input)

## Summary of changes

- routes all login through a global dialog
- refactors the dialog
  - changed from `AlertDialog` to regular `Dialog` component
  - adds ability to sign in with magic links
  - adds a `NEXT_PUBLIC_MAGIC_LINK_ENABLED` flag

## Testing

Preview branch: https://preview-5521-auth-flow.dc.rdc-staging.library.northwestern.edu/

locally:

```bash
export NEXT_PUBLIC_MAGIC_LINK_ENABLED=true
npm run dev
```

## Notes

The intent of `NEXT_PUBLIC_MAGIC_LINK_ENABLED` is to control whether or not Magic Links should be enabled at the app level. This provides, for the moment, an easy kill switch for the feature. If the environment variable is `true`, then we also check the new endpoint.


components/Shared/AuthDialog.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    render(withUserProvider(<AuthDialog />));
+
+    const magicLinkSection = screen.queryByTestId("magic-link");
+
+    expect(magicLinkSection).not.toBeInTheDocument();
+  });
+
+  describe("Magic Link enabled", () => {
+    beforeEach(() => {
+      process.env.NEXT_PUBLIC_MAGIC_LINK_ENABLED = "true";
```
kdid, line 66: @charlesLoder I'm questioning why we need this variable. If our overall goal is to remove as much logic as possible from the front end, if this is needed we should add `/capabilities/magic/login` route to the API. But is that even necessary at this point?

charlesLoder: It isn't entirely needed, but I think it can serve a purpose.

I'm imagining a scenario where it is requested we immediately remove the magic link functionality because of reasons from on high. We could simply change the env variable in Amplify and redeploy, which is faster than updating the api. Maybe that isn't a scenario that would happen though

mbklein: I have the same question about the need for the toggle variable – changing the response from the `/capabilities/magic/login` route in the API is an easy and immediate toggle, without having to redeploy anything, so I think it'd be simpler to leave it there.

charlesLoder: That works for me

mathewjordan (approved):
This is so nice. I'm interested in @davidschober  and @vlsrobinson perspective on the modal layout/language. I think there could be some additional guidance there on the perks for being logged in but that I think can be incremented on. Overall it is awesome and really ties together things.

mathewjordan:
Followup: I just noted that this hasn't been rebased on the scopes work by @kdid in #424. We may want to verify that everything comes in cleanly without conflict.

components/Shared/AuthDialog.test.tsx
charlesLoder: var removed and tests updated

charlesLoder:
@mathewjordan — rebased

@kdid — placeholder and aria added (Lastpass kept adding my email so I never noticed the missing placeholder!)

mathewjordan:
Nice. Seeing those beautiful locks now.

image

mbklein (approved):

PR #426, opened 2025-05-14
Add share work functionality via content state.

mathewjordan:
## What does this do?

This Adds in the ability for a user to share the current IIIF Content State of a the Clover IIIF Viewer on a work page. To implement this, we have updated Clover IIIF to `latest` and use the `contentStateCallback` to track to state. A new **Share** button has been added adjacent to to the **View as IIIF...** button. This opens up our dialog that we already use for **Download and Share** further down the screen. You'll notice that this page has been reorganized as an accordion list for greater legibility. The first box/section is the Share URL form which allows a user to construct a Share URL.

### Anatomy of a Share URL

- Defaults to the current URL `/items/{work.id}`
- If user checks **Jump to current image** (`i` of `total`), then a `?iiif-content` param is appended to the URL constructing a IIIF Content state that sets the active canvas for this particular work
- If user checks **Zoom to current view of the image** OR **Start at current time**, then a `?iiif-content` param is modified further, containing a respective `xywh` or `t` selector
- If user inputs an annotation text, then then a `?iiif-content` param is modified further with a TextualBody value for the annotations `body`

The user can copy this URL at at any point and share. The link will bring the user to the current page,. If a `iiif-content` param is present, then the work page at `pages/items/[id].tsx` will parse this value and hand it back to the Viewer. At this point, Clover will handle the parsing of the encoded content state string.

## How do we review?

Please review in your dev or at https://preview-5525-share-content-state.dc.rdc-staging.library.northwestern.edu/

- Go to any work, preferably one with multiple access filesets.
- Go to fileset that is not the first item
- If an Image work, click on the IIIF Viewer to zoom/pan. If Video/Audio, scan to a new time in the media file
- Click Share
- Experiment with various shares of this item
  * Share URL as-is (not to current fileset, no zoom)
  * Share URL as current fileset
  * Share URL as current fileset and zoomed
  * Share URL as current fileset, zoomed, and some textual annotation body
- Try this with various works

image

image

image

## Notes

Apologies in advance for the high number of files this touches. Please let me know if you have questions. This functionality was difficult to implement without some reworking of our work pages dialogs.


mbklein (approved):
This is great. I love how the annotations render when the page loads. 👍🏻

kdid (approved):
Looks 👍 to me.

mathewjordan:
Leaving branch up in the near-term in case there are trailing reviews.

charlesLoder:
@mathewjordan my only thought — should the information toggle be available so users can toggle off annotations? Absolutely love this feature though

mathewjordan:
> @mathewjordan my only thought — should the information toggle be available so users can toggle off annotations? Absolutely love this feature though

I agree! This would be awesome feature. However - will need to be introduced in the Clover side of things.

PR #450, opened 2025-07-31
Update all search results messages to include filters of facets

mathewjordan:
## What does this do?

This work abstracts out search results message under a single component and a set of helpers. In doing so, we are now appending any applied facets to the message. This can be reviewed at https://preview-5606-unify-results-message.dc.rdc-staging.library.northwestern.edu/search

### Legacy search example

On a legacy search, counts will render. For example, the message for a search on "joan baez" with _genre_ filtered by **black-and-white negatives** will render as:

```
469 results for “joan baez” filtered by genre for black-and-white negatives
```

https://preview-5606-unify-results-message.dc.rdc-staging.library.northwestern.edu/search?q=joan+baez&genre=black-and-white+negatives

image

### AI Mode search tool returns

On a AI Mode search tool, the designated `message.input.query` will be the term searched for and render. So if the a question for “joan baez” prompts the LLM searches for "joan baez photographs" the returned message would be:

```
Results for “joan baez photos” filtered by genre for black-and-white negatives
```

Clicking **View Results** would bring the user to a search results screen for those results where they will see the message again.

image

### AI Mode stack tooltips

On AI Mode, whenever a "Stack" of works with a tooltip is present, a message will display as well.

image

### DIFF
--- a/components/Chat/Response/Interstitial.styled.tsx
+++ b/components/Chat/Response/Interstitial.styled.tsx
@@ -1,4 +1,4 @@
-import { keyframes, styled } from "@/stitches.config";
+import { styled } from "@/stitches.config";
 
 const StyledInterstitialIcon = styled("div", {
   display: "flex",
@@ -32,6 +32,7 @@ const StyledInterstitialWrapper = styled("div", {
   display: "flex",
   justifyContent: "space-between",
   alignItems: "center",
+  marginBottom: "$gr1",
 });
 
 const StyledInterstitial = styled("div", {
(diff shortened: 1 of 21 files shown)


bmquinn:
I like the way these look, much clearer about what is happening in the current context. I noticed two things reviewing the preview (second of which may be a separate issue too):
- If you have AI unchecked it says you're searching for "undefined", which we might want to catch as a special case?

undefined

- When you uncheck "AI Mode" the context doesn't clear out even when you refresh because of the URL params:
unchecked_ai_mode

vlsrobinson (approved):

kdid (approved):
So, I do see that if I click the link you put in the issue and land on the regular search mode I see the counts and the facets/filters in the results message. One thing that hung me up:

- I kept clicking the link but I was already in AI mode so I just kept landing on a blank page and it seemed like it wasn't working. 

Approving because I don't think that's what you are addressing here.

PR #453, opened 2025-08-05
Keep facets and carry on.

mathewjordan:
## What does this do?

This work will keep the the filtered context of a conversation going if it exists and carry it over to the followup textarea until dismissed or replaced by and new filter. Much of the work included here is part of the process to clean up the passing of context between an initial search, rendered interstitials, the results panel where filters can be refined, and the conversations followup textarea. Many files are touched with all of these moving parts.

### Summary

- Filters will carry through to conversation followup textarea if Collection is searched within
- Filters will carry over from the results panel when use clicks back
- Filters will be dismissed if user manually dismisses 

### Notes

- To avoid rerenders of search results, a new layout-context has been created to separate search logic watching if the search bar is fixed.

## How should we review?

https://preview-5597-keep-facets-and-carry-on.dc.rdc-staging.library.northwestern.edu/

- Go the preview branch.
- Start a search from a collection.
- See that facets carry over.
- Click on a View results -> interstitial and refine
- See that the refined facets carry over
- Ask a followup (note you may need to be explicit to the LLM about the changes in your refinement as it may still be thinking you care more about the original context)
- Dismiss context

### DIFF
--- a/components/Chat/Chat.test.tsx
+++ b/components/Chat/Chat.test.tsx
@@ -22,7 +22,6 @@ jest.mock("@/context/search-context", () => {
           body: [],
           ref: "",
         },
-        searchFixed: false,
       },
     }),
   };
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -35,10 +35,6 @@ const Chat = () => {
               aggregations: [],
             },
           ],
-          /**
-           * Clear the staged context after submitting a question
-           */
-          stagedContext: undefined, // clear chat context on new question
         },
       });
     }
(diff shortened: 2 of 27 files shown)


mbklein:
Once the user returns from refining the facets, is there any visual indication of what facets/filters are active in the conversation view? I was able to confirm that the additional facets I selected were still active by entering a followup question that I knew would generate an interstitial with the facet details. It might be nice to indicate the current context right above the followup question field.

mathewjordan:
> Once the user returns from refining the facets, is there any visual indication of what facets/filters are active in the conversation view? I was able to confirm that the additional facets I selected were still active by entering a followup question that I knew would generate an interstitial with the facet details. It might be nice to indicate the current context right above the followup question field.

The only indicator right now is the thumbnail "Stack" with a tooltip. Adding another bit of text was something we explored in a previous issue but I think we opted to not go down that route yet.

PR #466, opened 2025-12-12
Fix similar slider in non AI mode

kdid:
### Summary

Fix `404` error in the "View All" link on the "More Like This" slider on work pages (non-AI mode). 

### Changes in this PR

- The `buildFacetFilters` function in `facets.ts` was trying to look up `similar` in the facets list, but similar isn't a facet, it's a special query parameter that's handled separately in the query builder as a `more_like_this` query. Added a check to skip the `similar` parameter in `buildFacetFilters` so it doesn't interfere with the filter building.
- In `chat-helpers.js` added a type check to ensure `facet[facetId]` is a string before calling `.replace()`. If it's not a string, it defaults to an empty string.

Screenshot 2025-12-12 at 8 33 03 AM

### DIFF
--- a/lib/chat-helpers.ts
+++ b/lib/chat-helpers.ts
@@ -79,7 +79,9 @@ function createResultsMessageFromContext(
 
     const facetField = getFacetIdByField(facetId);
     // @ts-ignore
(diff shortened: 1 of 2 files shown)


mathewjordan (approved):
💯

PR #474, opened 2026-04-30
Implement content search within transcriptions

kdid:
## Summary

When a keyword search is performed, the app now concurrently queries a fileset search endpoint for phrase matches within transcription annotations. If content matches are found, a view toggle appears in the results header letting users switch between standard item results and a "within content" view.

The content view displays matches grouped by work - each group shows the work thumbnail alongside a list of matching file sets, each with a thumbnail, a link to the specific canvas, and a snippet of the matching transcription text with the search term highlighted. Pagination is handled independently from the main results. Filters are suppressed in the content view with an explanatory note; the visibility facet is applied to the fileset query itself.

## New files

- `lib/queries/fileset-search.ts` - builds the fileset search request body
- `lib/utils/fileset-helpers.ts` - groups fileset results by work
- `components/Search/TranscriptionResults.tsx` - renders grouped content matches
- `components/Search/TranscriptionResults.styled.ts` - styles for the above
- `components/Search/TranscriptionPagination.tsx` - pagination for content results


bmquinn (approved):

mathewjordan (approved):
Looks great and love how much it adds.

PR #486, opened 2026-06-16
Add app version in footer

kdid:
## Summary

Adds a version indicator to the site footer, similar to how Meadow displays its version. The display very unobtrusive and is environment-aware to avoid showing links to releases that don't exist yet.

### Changes

- **Production**: displays the version as a link to the GitHub release tag (e.g. `v3.1.4` -> `github.com/nulib/dc-nextjs/releases/tag/v3.1.4`)
- **Staging**: displays the version as plain text with the short commit SHA linking to the commit on GitHub (e.g. `v3.1.4 (staging abc1234)`)
- **Dev**: displays the version as plain text with the short commit SHA if available (e.g. `v3.1.4 (DEV abc1234)`)
- Version is read from `package.json`; environment is read from the existing `HONEYBADGER_ENV` and `HONEYBADGER_REVISION` env vars already exposed via `next.config.js`

Screenshot 2026-06-16 at 9 09 01 AM

### DIFF
--- a/components/Footer/Content.tsx
+++ b/components/Footer/Content.tsx
@@ -8,13 +8,12 @@ import {
   FooterVersion,
   Social,
 } from "./Footer.styled";
-import { version } from "@/package.json";
+import { getBuildInfo } from "@/lib/build-info";
 
 import React from "react";
 
 export const FooterContent: React.FC = () => {
-  const deployEnv = process.env.HONEYBADGER_ENV;
-  const revision = process.env.HONEYBADGER_REVISION;
+  const { version, releaseUrl, commitUrl, shortSha, label } = getBuildInfo();
 
   const currentYear = () => {
     const today = new Date();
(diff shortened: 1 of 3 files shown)


bmquinn (commented):
Could we move the env logic into a small app-level helper like `lib/build-info.ts` with a `getBuildInfo()` function? The helper could normalize the deployment cases: production gets a release tag link, staging gets the version plus linked short SHA, and every other env gets a non-production label plus the short SHA when available.

That would keep the Honeybadger-specific env names out of the footer and let the component render from the helper instead of deciding the behavior itself. Something like this maybe?
```js
const buildInfo = getBuildInfo();

<FooterVersion as={buildInfo.href ? "a" : "span"} href={buildInfo.href}>
  {buildInfo.label}
  {buildInfo.sha && buildInfo.shaHref && (
    <>
      {" "}
      (<a href={buildInfo.shaHref}>{buildInfo.sha}</a>)
    </>
  )}
</FooterVersion>
```

bmquinn (approved):
Looks great 👍

PR #488, opened 2026-06-26
Bugfix - public items appearing as institution

kdid:
## Summary

Users reported that sometimes 'Public' assets are shown as restricted (thumbnails have the lock icon and the "Authentication Needed..." banner is displayed). A page refresh fixes the error. 

The root cause is a race condition: `UserContext` has no loading state. `user` starts as `null` and `getUser()` is async. Every component that checks `user?.scopes?.includes('read:Public')` sees `null` before the request completes and evaluates to `false` → restricted. Once the auth call resolves and user scopes are populated, re-renders fix it. That's why a refresh "helps"  timing varies. The race window is very short - just the time it takes for `/auth/whoami` to respond. By the time you type a URL and the page visually settles in front of you, the auth call has almost certainly already completed and the UI has already self-corrected. You'd have to be watching the exact millisecond of first paint to catch it.

The users who see it are likely experiencing it as a visible flash or a stuck state — depending on how fast their network resolves that auth endpoint. There's a range of possibilities:
- Fast connection: brief flicker of lock icons that self-corrects quickly — easy to miss
- Slower connection or higher latency to `/auth/whoami`: the restricted state persists long enough to notice, take a screenshot, or share the link while the page is still in that state
- If they shared the link while the page was already rendered in the restricted state in their browser, your loading that same URL in a fresh context would start the race fresh — and again resolve before you see it

The fix: add `isLoading` to `UserContext`, start it `true`, and set it `false` only after `getUser()` settles. Then guard all "show as restricted" conditions behind `!isLoading`.

fixes https://github.com/nulib/repodev_planning_and_docs/issues/5826

## Changes

**Core fix (5 files):**
- `types/context/user.ts` — added `isLoading: boolean` to the `UserContext` type
- `context/user-context.tsx` — added `isLoading` state starting at `true`; moved `setIsLoading(false)` to fire after `getUser()` settles whether or not a user was found
- `hooks/useWorkAuth.ts` — reads and returns `isAuthLoading` from the context
- `components/Grid/Item.tsx` — `isRestricted` returns `false` while loading
- `components/Search/TranscriptionResults.tsx` — same guard on `isRestricted`

**Work display (4 files):**
- `pages/items/[...id].tsx` — `WorkRestrictedDisplay` now requires `!isAuthLoading` in addition to `!userCanRead`
- `pages/embedded-viewer/[manifestId].tsx` — passes `isAuthLoading` down to `EmbeddedViewer`
- `components/Work/EmbeddedViewer.tsx` — accepts `isAuthLoading` prop; suppresses `WorkRestrictedDisplay` while loading
- `components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx` — suppresses the "Download requires NetID" announcement while loading

**Test mocks (5 files):** Added the now-required `isLoading: false` (plus pre-existing missing `scopes`, `isInstitution`, and modal stubs that were already type errors).

## Testing

- Not sure how to test, I think this might be a deploy and see if the issue re-occurs situation. 
- But we should  check it didn't introduce unintended regressions

### DIFF
--- a/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
+++ b/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
@@ -48,6 +48,9 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
   const [imageCanvases, setImageCanvases] = React.useState<Canvas[]>([]);
   const { userCanRead, isAuthLoading } = useWorkAuth(work);
   const isSharedLinkPage = router.pathname.includes("/shared");
+  const loginRequired =
+    (isAuthLoading && work?.visibility !== "Public") || // Work is not public and auth is loading
+    (work && !userCanRead && !isSharedLinkPage); // User cannot read and it's not a shared link page
 
   React.useEffect(() => {
     if (manifest?.items && Array.isArray(manifest?.items)) {
@@ -62,7 +65,7 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
     <EmbedResourcesWrapper>
       <Heading as="h3">Download and Embed</Heading>
 
-      {!userCanRead && !isAuthLoading && !isSharedLinkPage && (
+      {loginRequired && (
         <Announcement>
           Download requires Northwestern University NetID authentication{" "}
         </Announcement>
--- a/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
+++ b/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
@@ -46,7 +46,7 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
 }) => {
   const router = useRouter();
   const [imageCanvases, setImageCanvases] = React.useState<Canvas[]>([]);
-  const { userCanRead } = useWorkAuth(work);
+  const { userCanRead, isAuthLoading } = useWorkAuth(work);
   const isSharedLinkPage = router.pathname.includes("/shared");
 
   React.useEffect(() => {
@@ -62,7 +62,7 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
     <EmbedResourcesWrapper>
       <Heading as="h3">Download and Embed</Heading>
 
-      {!userCanRead && !isSharedLinkPage && (
+      {!userCanRead && !isAuthLoading && !isSharedLinkPage && (
         <Announcement>
           Download requires Northwestern University NetID authentication{" "}
         </Announcement>
--- a/components/Work/EmbeddedViewer.tsx
+++ b/components/Work/EmbeddedViewer.tsx
@@ -8,10 +8,12 @@ import WorkViewerWrapper from "@/components/Clover/ViewerWrapper";
 const EmbeddedViewer = ({
   work,
   userCanRead,
+  isAuthLoading,
   searchParams,
 }: {
   work: Work;
   userCanRead?: boolean;
+  isAuthLoading?: boolean;
   searchParams: { [key: string]: string };
 }) => {
   const thumbnail = work?.thumbnail || "";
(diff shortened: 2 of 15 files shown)


pages/items/[...id].tsx
```diff
@@ -144,7 +145,7 @@ const WorkPage: NextPage<WorkPageProps> = ({
                   searchQuery={searchQuery}
                 />
               )}
-              {work && !userCanRead && (
+              {work && !userCanRead && !isAuthLoading && (
```
Comment (mbklein) [inline pages/items/[...id].tsx]:
```suggestion
              {(showRestricted || (work && !userCanRead)) && (
```

components/Work/EmbeddedViewer.tsx
```diff
@@ -48,7 +50,7 @@ const EmbeddedViewer = ({
           viewerOptions={viewerOptions}
         />
       ) : (
-        <WorkRestrictedDisplay thumbnail={thumbnail} />
+        !isAuthLoading && <WorkRestrictedDisplay thumbnail={thumbnail} />
```
mbklein:
```suggestion
        showRestricted && <WorkRestrictedDisplay thumbnail={thumbnail} />
```

components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
```diff
@@ -62,7 +62,7 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
     <EmbedResourcesWrapper>
       <Heading as="h3">Download and Embed</Heading>
 
-      {!userCanRead && !isSharedLinkPage && (
+      {!userCanRead && !isAuthLoading && !isSharedLinkPage && (
```
mbklein:
```suggestion
      {(showRestricted || (!userCanRead && !isSharedLinkPage)) && (
```

components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
```diff
@@ -46,7 +46,7 @@ const EmbedResources: React.FC<EmbedResourcesProps> = ({
 }) => {
   const router = useRouter();
   const [imageCanvases, setImageCanvases] = React.useState<Canvas[]>([]);
-  const { userCanRead } = useWorkAuth(work);
+  const { userCanRead, isAuthLoading } = useWorkAuth(work);
   const isSharedLinkPage = router.pathname.includes("/shared");
```
mbklein, line 54:
```suggestion
  const showRestricted = isAuthLoading && work?.visibility !== "Public";
  
```

components/Work/EmbeddedViewer.tsx
mbklein, line 44:
```suggestion

  const showRestricted = isAuthLoading && work?.visibility !== "Public";
  
```

pages/items/[...id].tsx
```diff
@@ -48,7 +48,8 @@ const WorkPage: NextPage<WorkPageProps> = ({
   const [initialLabel, setInitialLabel] = useState<string | undefined>();
   const [initialSnippet, setInitialSnippet] = useState<string | undefined>();
   const canvasParamProcessed = useRef(false);
-  const { userCanRead, isWorkReadingRoomOnly } = useWorkAuth(work);
+  const { userCanRead, isWorkReadingRoomOnly, isAuthLoading } =
+    useWorkAuth(work);
   const router = useRouter();
   const { isChecked: isAI } = useGenerativeAISearchToggle();
```
Comment (mbklein) [inline pages/items/[...id].tsx]:
```suggestion
  const showRestricted = isAuthLoading && work?.visibility !== "Public";
```

mbklein (commented):
Updated restricted logic in a few more places.

components/Work/EmbeddedViewer.tsx
kdid: I like the idea of extracting these. But have we lost a case here? 

If we replace `!isAuthLoading && <WorkRestrictedDisplay>` with just `showRestricted && <WorkRestrictedDisplay>`, where `showRestricted = isAuthLoading && work?.visibility !== "Public"`. After auth finishes loading, `isAuthLoading` is `false`, so `showRestricted` is always `false` - the restricted display would never render for users who genuinely can't access the work. (Or am I getting the flow wrong?)

mbklein: Yeah, I had a feeling I dropped some and/or logic along the way there, and I think you've got it right on this one.

PR #489, opened 2026-06-29
Fix lock icons with visibility check

bmquinn:
## Summary

Fixes intermittent lock icons appearing on public search result thumbnails by checking item visibility before user scope state. Public items now render as unrestricted even if auth resolves without a user, while non-public items still respect loading state and matching read scopes.

## Changes

- Checks for visibility explicitly in grid items before user scope

## Testing

- https://preview-fix-public-lock-icons.dc.rdc-staging.library.northwestern.edu/

### DIFF
--- a/components/Grid/Item.test.tsx
+++ b/components/Grid/Item.test.tsx
@@ -1,6 +1,9 @@
 import { render, screen } from "@/test-utils";
 import GridItem from "./Item";
+import React from "react";
 import { SearchShape } from "@/types/api/response";
+import { UserContext } from "@/context/user-context";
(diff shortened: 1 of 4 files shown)


kdid:
Good catch, let's fix this in transcription results too: https://github.com/nulib/dc-nextjs/blob/deploy/staging/components/Search/TranscriptionResults.tsx#L83

bmquinn:
@kdid if you search for "chicago" in the deployed preview branch there's results with mixed visibility for testing:
SCR-20260630-jjsc

kdid (approved):

PR #492, opened 2026-08-03
Prevent allow upscaling error for download resources

kdid:
## Summary

Fixes "Requested size requires upscaling" errors when clicking Download JPG from the Download and Embed tab on the shared-link and work pages, for images narrower than 3000px. Was also affecting the embed resources code.

fixes https://github.com/nulib/repodev_planning_and_docs/issues/5851

## Changes

- For the download, cap whichever dimension (width or height) is longer at 3000px, instead of always constraining width — a narrow-but-tall image (e.g. 1200×15000) now downloads/embeds at a capped height, not its full native height. Add allow upscaling to the IIIF request as a fallback if dimensions are not known.
- For the embed, dropped the incorrect percentage labels (1800px was mislabeled "50%" when it's actually 60% of 3000, and the percentages were never meaningful relative to the real image size anyway) in favor of plain pixel labels.
- Updated the one Playwright assertion (tests/work.spec.ts:542) that checked the old "...100%...50%" label text; the URL-format assertions didn't need changes since that fixture image is landscape (4032×3024, confirmed via its info.json).

## Testing

- Verify: Download JPG on a narrow <3000px-wide image (the request to the IIIF server should now reflect the image's actual width)
- Verify: Download working as expected for "regular" images
- Verify: Embed options working as expected

### DIFF
--- a/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
+++ b/components/Work/ActionsDialog/DownloadAndShare/EmbedResources.tsx
@@ -120,7 +120,6 @@ interface ItemProps {
 const Item: React.FC<ItemProps> = ({ item, showEmbedWarning }) => {
   const [embedHTMLOpen, setEmbedHTMLOpen] = React.useState(false);
   const [color, setColor] = React.useState("default");
-  const [width, setWidth] = React.useState(3000);
 
   const iiifImageInfo = `${getInfoResponse(item)}/info.json`;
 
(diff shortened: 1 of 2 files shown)


mbklein:
Good fix, but I have a question about `longestDimension`. There are several short circuits where `longestDimension` is only used if it has a value. But I can't see a code path where it _wouldn't_ end up with a value by the time it reaches those checks. Line 144 is pretty clear on setting it to either the height or the width. Am I missing something, or is it possible to clean up the logic by knowing that `longestDimension` is definitely going to be defined before those checks?

mbklein (approved):

PR #510, opened 2026-09-23
Work/Field level AI disclosure

kdid:
## Summary

**_**Note:** This is intended as a first iteration. It replaces the auto-generated AI notes and adds field-level detail to show only what the provenance data already supports. Both the granularity of the disclosure and where the logic lives are expected to change as we settle how much we want to surface._**

Replaces the free-text AI disclosure note with a structured indicator driven by the API's `ai_involved` / `ai_provenance` fields.

Previously, a work with AI-generated metadata carried a Local Note in descriptive metadata — *"Some metadata created with the assistance of AI (us.anthropic.claude-sonnet-4-6) on 2026-09-21"* — which rendered as one more row in the Notes list. It was easy to miss, unstructured, and didn't say which fields were affected.

Now a pill sits above the descriptive metadata list reading **"AI-assisted metadata"**, **"AI-assisted transcriptions"**, or both. An ⓘ tooltip lists the affected fields and transcriptions, each with a short status:

> **Descriptive Metadata fields with AI involvement:**
> Alternate Title: AI generated, human reviewed
> Creator: AI generated, human reviewed
> Description: AI + human edited
> Style Period: AI generated, human reviewed
>
> **This item includes 2 AI-assisted transcriptions:**
> p. 1 verso: AI + human edited
> p. 1 recto: AI generated

The pill renders outside the `Expand` wrapper so it can't be hidden by the "Show More" collapse.

## Testing

Preview branch: https://preview-ai-disclosure.dc.rdc-staging.library.northwestern.edu/

Examples:
- Meatadata only: https://preview-ai-disclosure.dc.rdc-staging.library.northwestern.edu/items/46f991a6-cb61-4c89-9a6f-779c6ef86486
- Transcriptions only: https://preview-ai-disclosure.dc.rdc-staging.library.northwestern.edu/items/b0e7d25b-cd66-4bda-8dec-a1b462318257
- Both metadata and transcriptions: https://preview-ai-disclosure.dc.rdc-staging.library.northwestern.edu/items/8eaa818f-989d-44b7-9c6c-1a57e766d081

Note: The legacy Local Notes still render in the Notes row while they exist in the data. Removing them is a Meadow-side change; no filtering was added here.

## Relationship to Meadow

This consumes `Provenance.work_summary_map/1` as exposed by dc-api-v2, and reuses Meadow's vocabularies rather than inventing parallel ones. Three are copied deliberately — **if they change in Meadow, they need to change here:**

| This repo (`lib/ai-provenance.ts`) | Source of truth in Meadow |
| --- | --- |
| `AI_ORIGINS` allowlist | `@ai_involved_origins` — `lib/meadow/ai/provenance.ex` |
| `ORIGIN_LABELS` (verbatim) | `ORIGIN_META` — `components/AIProvenance/Badges.jsx` |
| `INACTIVE_STATUSES` | `INACTIVE_FIELD_STATUSES` — `Badges.jsx` |

Consequences of using the allowlist: `human_generated` and `human_or_legacy` (the nil-origin fallback at `provenance.ex:1145`) are never disclosed as AI. Non-`applied` entries — `proposed`, `reviewed`, `rejected`, `failed`, `deleted` — are filtered out, since the recorded value isn't the live one. `work_summary_map` doesn't filter by status, so this is a guard on our side; worth confirming whether the indexer filters upstream.

**One deliberate divergence.** Meadow badges origin alone, so `ai_generated` + `human_reviewed` reads "AI generated". We append review: **"AI generated, human reviewed"**. That a human checked AI-generated content is a primary disclosure for a public audience, and its absence is meaningful too.

This is narrow on purpose. `human_oversight_level` is derived from `origin` for every value except one — see `human_oversight_for_origin/1` vs `review_target!/5`, which sets `human_reviewed` only on actual approval. So it's the single oversight value carrying information the origin label doesn't already have, and the only one we append. Origin remains the base label everywhere else, which keeps `ai_assisted_human_modified` reading "AI + human edited" rather than collapsing to "Human edited".

Nothing staff-facing is public: no reviewer, no model, no dates, no PREMIS or C2PA. Those are modelled in the types so they're cheap to surface later, but nothing renders them.

## What this repo does on its own

- **Field key → display label.** DC renders metadata from the IIIF manifest, whose labels don't match API field names, so `date_created` → "Date", `library_unit` → "Department", with a snake_case → Title Case fallback. Note `description` has no metadata row — it's the summary paragraph under the title — but it's still named in the tooltip.
- **Transcription → file set label.** `file_set_annotations` keys are `content:<FileSetAnnotation id>`, and those ids don't appear anywhere in the work payload, so a transcription can't be attributed to a page from `ai_provenance` alone. `getTranscriptionFileSetLabels` makes one `/search/file-sets` query per work to map annotation id → file set label, lazily on first tooltip open. If it fails, the count sentence still renders without the list.
- **`ai_involved` handling.** The two section flags come from `ai_involved` when
  the API supplies them, and are computed from `ai_provenance` when it doesn't.
  The check is per-key and tests for an actual boolean, so an explicit `false` is
  honored (it suppresses the indicator even when `ai_provenance` has AI entries)
  and a half-populated object still works. `ai_involved` isn't in the API yet, so
  every work currently takes the derived path; it will switch over on its own
  once the field ships.n present, derived from `ai_provenance` when absent — it isn't in the API yet, so this works before and after it ships.
- **Local types.** `@nulib/dcapi-types@2.9.1` has neither key. `Work` is a type alias (`components["schemas"]["Work"]`) and aliases can't be declaration-merged, so `types/api/ai-provenance.ts` defines them locally and intersects. Deletable as a unit once the package ships them.

## Changes

- Add `lib/ai-provenance.ts` — origin/status filtering, status and field labels, file set label lookup
- Add `components/Work/AIProvenance.tsx` + styles — pill and tooltip, reusing the Radix tooltip pattern from `GenerativeAIToggle`
- Add `types/api/ai-provenance.ts` — `ai_involved` / `ai_provenance` types, incl. PREMIS and C2PA (modelled, not rendered)
- Render the indicator in `TopInfo.tsx` inside a new `TopInfoMetadata` wrapper, keeping `TopInfoContent`'s two-column grid intact
- Move `TooltipTrigger` / `TooltipContent` from `Search/GenerativeAI.styled.ts` to `Shared/Tooltip.styled.tsx` — Chat already reached into Search for them, Work would have been the third; both consumers updated
- Copy lives in `lib/constants/common.tsx` so wording can be iterated without touching components
- ~60 unit tests covering the origin allowlist, status filtering, the review divergence, and label mapping

Screenshot 2026-09-21 at 9 50 42 AM


bmquinn (approved):
The tooltips are working great, I really like the dynamic text whether it's transcription, metadata, or both. 👏 For me, this is a crucial first step for getting something in front of users while we work on the broader C2PA functionality in an iterative way (and is slightly more community focused, as others are interested in that spec as well beyond NUL)

